PR #25, opened 2024-01-16
MCC DAQ Driver

franz-sweepMe:
Add driver for Logger MCC-DAQ


fk3:
Folder structure for the 3.9 libraries still needs to be fixed

fk3 (changes requested):
Folder structure for the 3.9 libraries still needs to be fixed

src/Logger-MCC_DAQ/main.py
```diff
@@ -0,0 +1,196 @@
...
+
+# this driver needs libraries installed by the manufacturer software InstaCal
+try:
+    from mcculw import ul
+    from mcculw.ul import ULError
+    from mcculw.device_info import DaqDeviceInfo
+    from mcculw.enums import ScanOptions, FunctionType, Status, AnalogInputMode, InterfaceType, ULRange
+except FileNotFoundError:
+    msg = "MCC DAQ Software missing. Install InstaCal and Universal Library (UL)."
+    raise Exception(msg)
```
fk3, line 47:
```suggestion
except FileNotFoundError as e:
    msg = "MCC DAQ Software missing. Install InstaCal and Universal Library (UL)."
    raise ImportError(msg) from e
```
to make the causal dependency clear

src/Logger-MCC_DAQ/main.py
```diff
@@ -0,0 +1,196 @@
+# This Device Class is published under the terms of the MIT License.
+# Required Third Party Libraries, which are included in the Device Class
+# package for convenience purposes, may have a different license. You can
+# find those in the corresponding folders or contact the maintainer.
+#
+# MIT License
+#
+# Copyright (c) 2023 SweepMe! GmbH (sweep-me.net)
```
afischer-sweepme, line 8: 2023-2024

src/Logger-MCC_DAQ/license.txt
```diff
@@ -0,0 +1,28 @@
+This Device Class is published under the terms of the MIT License.
+Required Third Party Libraries, which are included in the Device Class
+package for convenience purposes, may have a different license. You can
+find those in the corresponding folders or contact the maintainer.
+
+MIT License
+
+Copyright (c) 2023 SweepMe! GmbH (sweep-me.net)
```
afischer-sweepme, line 8: 2023-2024

src/Logger-MCC_DAQ/main.py
```diff
@@ -0,0 +1,196 @@
...
+            msg = "Selected port was not found. Please use 'Find ports' and check whether the port still exists."
+            raise Exception(msg)
+
+        ul.create_daq_device(self.board_num, self.descriptor)
+        ul.flash_led(self.board_num)
+
+    def disconnect(self):
+        ul.release_daq_device(self.board_num)
+
+    def initialize(self):
```
afischer-sweepme, line 146: All configuration related properties should be moved to semantic standard function "configure" because configure is called whenever a branch of the SweepMe! sequencer starts to be processed. If users have multiple branches with each having a Logger module that is differently configured for the instrument, it would not change anymore because "initialize" is only called once at the beginning.

Therefore, "initialize" only contains commands that need to be used only once at the beginning of each measurement such as the analog input check or the ai range check.

Here, it would make sense to test the driver once being used in two different branches because sometimes there are problems when dlls are involved as they might have exclusive handles.

src/Logger-MCC_DAQ/main.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __init__(self):
+
+        EmptyDevice.__init__(self)
+
+        self.shortname = "MCC-DAQ"  # short name will be shown in the sequencer
+        self.variables = []
+        self.units = []
+
+        self.port_manager = False
```
afischer-sweepme, line 69: Because the port manager is switched off by default, this line can be removed. I think it is a leftover from the driver that was contributed.

https://github.com/SweepMe/pysweepme/blob/ada2b1ecfb8bad095862eb78046ab9178ea30bf1/src/pysweepme/EmptyDeviceClass.py#L65

afischer-sweepme (changes requested):
Thanks for improving the driver. Here are a few comments.

afischer-sweepme (approved):

fk3 (approved):

PR #26, opened 2024-01-16
Import Driver Test

franz-sweepMe:
Add a dictionary of drivers that cannot be tested and the corresponding reason. Add a check if the driver test should be skipped.

### DIFF
--- a/tests/importability/import_driver.py
+++ b/tests/importability/import_driver.py
@@ -51,15 +51,16 @@ def is_compatible(driver_name: str) -> bool:
 
 # Driver name and reason for skipping
 SKIPPED_DRIVERS = {
-    "Logger-MCC_DAQ" : "Driver requires installed manufacturer software",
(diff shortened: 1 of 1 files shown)


tests/importability/import_driver.py
```diff
@@ -90,9 +110,14 @@ def import_driver(driver_name: str) -> None:
...
-else:
+
+if not is_compatible(driver_name):
     logging.debug(
         f"Skipped importing {driver_name} because it is not meant to be compatible with this python version.",
     )
+elif should_skip_driver(driver_name):
+    logging.debug(
+        f"Skipped importing {driver_name} because it cannot be tested.",
+    )
```
fk3, line 121: That message is somewhat superfluous, as the `should_skip_driver()` function already logs that information including the reason.

franz-sweepMe: Removed additional log

fk3 (approved):

PR #27, opened 2024-01-19
Template style

fk3:
- Update template driver to adhere to current style guides
- Add a minimal driver template


afischer-sweepme (changes requested):
Can you add the webpage to the license, that's all I have.

afischer-sweepme (approved):

PR #28, opened 2024-01-25
Arduino all pins update

franz-sweepMe:
Add parameters to choose specific channels for readout. Enable flexible range of channels for varying Arduino Boards.

### DIFF
--- a/src/Logger-Arduino_AllPins/main.py
+++ b/src/Logger-Arduino_AllPins/main.py
@@ -60,18 +60,18 @@ class Device(EmptyDevice):
             "baudrate": 115200,
         }
 
-        self.max_voltage = 5
+        self.max_voltage = 5.0
 
         self.unit_dict = {
             "Volt": "V",
-            "Bit": " ",
+            "Numerical": "",
         }
 
     def set_GUIparameter(self):
         return {
             "Digital channel": "2,3,4,5,6,7,8,9,10,11,12,13",
             "Analog channel": "0,1,2,3,4,5,6,7",
-            "Analog unit": ["Volt", "Bit"],
+            "Analog unit": ["Volt", "Numerical"],
             "Resolution in Bit": 10,
         }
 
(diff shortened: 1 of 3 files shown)


src/Logger-Arduino_AllPins/main.py
```diff
@@ -26,61 +26,103 @@
...
+        self.unit_dict = {
+            "Volt": "V",
+            "Bit": " ",
+        }
+
+    def set_GUIparameter(self):
+        return {
+            "Digital channel": "2,3,4,5,6,7,8,9,10,11,12,13",
+            "Analog channel": "0,1,2,3,4,5,6,7",
+            "Analog unit": ["Volt", "Bit"],
```
afischer-sweepme, line 74: this could be "V" instead of "Volt"

afischer-sweepme: ok, I see now that you make a translation to "V" later, so it is fine.
Maybe "Analog return type" instead of "Analog unit" with options "Voltage" and "Bit level" ?

src/Logger-Arduino_AllPins/main.py
```diff
@@ -26,61 +26,103 @@
...
+            "Resolution in Bit": 10,
+        }
+
+    def get_GUIparameter(self, parameter={}):
+        self.variables = []
+        self.units = []
+
+        if parameter["Digital channel"] != "":
+            for pin in parameter["Digital channel"].split(","):
```
afischer-sweepme, line 83: here, you could use replace(" ", "") to remove all whitespaces that might be added by the users to the string

src/Logger-Arduino_AllPins/main.py
```diff
@@ -26,61 +26,103 @@
...
+                self.variables.append("Digital %i" % int(pin))
+                self.units.append("")
+
+        if parameter["Analog channel"] != "":
+            for pin in parameter["Analog channel"].split(","):
```
afischer-sweepme, line 88: here, you could use replace(" ", "") to remove all whitespaces that might be added by the users to the string

src/Logger-Arduino_AllPins/main.py
afischer-sweepme: one could also remove or replace all "D" in the string in case people prefer to use "D2, D3, ...." which would be then supported as well.

src/Logger-Arduino_AllPins/main.py
afischer-sweepme: one could also remove or replace all "A" in the string in case people prefer to use "A0, A1, ...." which would be then supported as well.

src/Logger-Arduino_AllPins/main.py
```diff
@@ -26,61 +26,103 @@
...
+                self.variables.append("Analog %i" % int(pin))
+                self.units.append(self.unit_dict[parameter["Analog unit"]])
+
+        self.resolution = 2 ** int(parameter["Resolution in Bit"]) - 1
 
     def initialize(self):
-        
         # we only need to wait once to receive the setup message
         if self.port.port_properties["NrDevices"] == 0:
```
afischer-sweepme, line 96: The entire use of self.port.port_propertes["NrDevivces"] is a bit hacky solution that we do not use anymore. Could you change alle occurences to the use of self.device_communication?

src/Logger-Arduino_AllPins/main.py
```diff
@@ -26,61 +26,103 @@
...
-        for i in np.arange(8):
-            self.variables.append("Analog %i" % i)
-            self.units.append("")
-
-        self.plottype = [True for x in self.variables] # True to plot data
+
+        self.plottype = [True for x in self.variables]  # True to plot data
         self.savetype = [True for x in self.variables]  # True to save data
-        
-        #self.idlevalue = 0 # motor position to go home
```
afischer-sweepme, line 60: this line can be removed. self.idlevalue is not supported anymore

src/Logger-Arduino_AllPins/main.py
```diff
@@ -26,61 +26,103 @@
...
-                                 "EOL": "\n",
-                                 "baudrate":115200,
-                                }
+        self.port_properties = {
+            "timeout": 3,
+            "EOL": "\n",
+            "baudrate": 115200,
+        }
+
+        self.max_voltage = 5
```
afischer-sweepme, line 63: could be float

src/Logger-Arduino_AllPins/main.py
```diff
@@ -26,61 +26,103 @@
...
+
+        self.unit_dict = {
+            "Volt": "V",
+            "Bit": " ",
```
afischer-sweepme, line 67: Empty unit can be defined "" i.e. an empty string

afischer-sweepme (changes requested):
Thanks for changing the driver. It looks already very nice and I just have some style related comments.

afischer-sweepme (approved):

PR #29, opened 2024-01-30
Arduino MCP4728

franz-sweepMe:
Add driver to control up to 8 MCP4728 boards using Arduino

### DIFF
--- a/src/Switch-Arduino_MCP4728/main.py
+++ b/src/Switch-Arduino_MCP4728/main.py
@@ -292,5 +292,5 @@ class Device(EmptyDevice):
         # Check Arduino response
         ret = self.port.read()
         if ret != f"ACK{gain}":
-            msg = f"Failed to set gaine (vref). Arduino response: {ret}"
+            msg = f"Failed to set gain (vref). Arduino response: {ret}"
             raise Exception(msg)
--- a/src/Switch-Arduino_MCP4728/main.py
+++ b/src/Switch-Arduino_MCP4728/main.py
@@ -35,14 +35,13 @@ from pysweepme import EmptyDevice
 
 class Device(EmptyDevice):
     description = """
-                        <h3>MCP4728</h3>
-                        <p>0x60 is standard. I2C Address is given either as 0-7 (HEX) or 0x60-0x67 (DEC)</p>
-                        If uncertain, you can use I2C scanner with arduino to check your I2C address
-                        Voltage Reference: Use either internal 2V, which can be amplified to internal 4V
-                        When using external reference voltage, set it
-                        Multi Channel Use: Parameter need to be set as string devided by ":"
-                        Multi Board use: only works with multi channel. Give list of number of boards * 4 inputs
-                        """
+        <h3>Arduino MCP4728</h3>
+        <p>This driver allows to set output voltages at MCP 4728 boards with 12-bit resolution. It can control up to 8 boards, each with 4 pins.</p>
+        <p>Set <em>Channel </em>to the pin number you want to set, or <em>all </em>to set all four channels. The voltage values must be passed as a colon-separated string: <em>1.0:2.5:0:4.2</em></p>
+        <p>The I&sup2;C address is set as integer 0-7, corresponding to the boards standard addresses 0x60-0x67 (HEX). You can check your devices' address by using an <a href="https://playground.arduino.cc/Main/I2cScanner/">I&sup2;C Scanner</a>.</p>
+        <p>The maximum voltage is defined by the Voltage reference, which can either be internal (2.048 V or 4.096 V by using 2x gain) or from an external source, e.g. the Arduino's 5 V or 3.3 V output. When using an external reference, the voltage must be given.</p>
+        <p>To use multiple MCPs, their I&sup2;C addresses need to be changed individually, as described <a href="https://github.com/jknipper/mcp4728_program_address">here</a>. Choose <em>Channel: all </em>and set a comma-separated list for <em>I2C Address: 0, 1, 2</em>.</p>
+    """
 
     def __init__(self) -> None:
         EmptyDevice.__init__(self)
--- a/src/Switch-Arduino_MCP4728/main.py
+++ b/src/Switch-Arduino_MCP4728/main.py
@@ -1,0 +1,271 @@
+# This Device Class is published under the terms of the MIT License.
+# Required Third Party Libraries, which are included in the Device Class
+# package for convenience purposes, may have a different license. You can
+# find those in the corresponding folders or contact the maintainer.
+#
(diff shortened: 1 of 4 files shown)


src/Switch-Arduino_MCP4728/main.py
```diff
@@ -0,0 +1,269 @@
...
+    """ here, semantic standard functions start that are called by SweepMe! during a measurement """
+
+    def connect(self):
+        # Set Name/Number of COM Port as key
+        instance_key = f"mcp4728_{self.port.port.port}"
+
+        if instance_key not in self.device_communication:
+            # Wait for Arduino initialization
+            self.port.read()
+            self.device_communication[instance_key] = "Connected"
```
fk3, line 131: Don't we need to unset the device communication flag somewhere when the measurement ends?

And on another note, as the comment talks about the initialization of the Arduino, this code could as well go into the initialize() function of the driver.

Finally a question of understanding: I can only see the code that it is waiting for the Arduino to finish initialization. But at which point is the arduino told to turn on or initialize?

franz-sweepMe: I've updated the instance_key parameter and added the disconnect function to pop the communication flag.

As far as I know, the Arduino initializes as soon as the connection is established and runs its setup() function.

afischer-sweepme: The reason why I think the code should be in 'connect' is that the semantic meaning of 'connect' is that a driver must be ready to communicate afterwards. As the Arduino is just ready for communication after we receiver the initial message, it should be done in 'connect'. This affects also the use with pysweepme, where one needs to use the connect method once to make sure the the Arduino is ready.
In future, get_driver of pyssweepme could automatically call 'connect' as this is always needed to be able to start a communication.
The semantic function "initialize" is more meant to be used to setup an instrument to the right configuration.

afischer-sweepme: Unsetting the flag in self.device_communication is not needed because there is not further decision at the end of a measurement whether the key is in or not. At the beginning of the next measurement, self.device_communication is anyway emptied, so I would suggest to remove the removal of the indentifier key at the end of a measurement from the code as it would otherwise suggest it has any meaning and could be rather confusing for people that take the driver as example.

fk3: in SweepMe! maybe, but what if someone wants to use the driver in pysweepme? Then they have to restart their application every time they want to re-establish the connection

src/Switch-Arduino_MCP4728/main.py
```diff
@@ -0,0 +1,269 @@
...
+            "Voltage in V": "V",
+            "Output in %": "%",
+        }
+
+    def set_GUIparameter(self):
+        return {
+            "Channel": ["0", "1", "2", "3", "all"],
+            "I2C Address": "0",
+            "SweepMode": ["Voltage in V", "Output in %"],
+            "Voltage reference": ["External", "Internal 2.048 V", "Internal 4.096 V"],
```
afischer-sweepme, line 76: I suggest to put Internal 2.048 V at first place and External at the last place. If no external voltage is connected, it could otherwise float. So people that use External have to change it. Otherwise, 2.048 at max are applied

src/Switch-Arduino_MCP4728/main.py
```diff
@@ -0,0 +1,269 @@
...
+# Device: Arduino MCP4728
+
+
+from pysweepme import EmptyDevice
+
+
+class Device(EmptyDevice):
+    description = """
+        <h3>Arduino MCP4728</h3>
+        <p>This driver allows to set output voltages at MCP 4728 boards with 12-bit resolution. It can control up to 8 boards, each with 4 pins.</p>
```
afischer-sweepme, line 39: Here, I would add that the driver only works if the .ini file shipped with the driver is loaded onto the Arduino. Further, additional libraries that must be downloaded in the Arduino IDE can be mentioned here.

src/Switch-Arduino_MCP4728/main.py
```diff
@@ -0,0 +1,269 @@
...
+
+        self.shortname = "Arduino MCP4728"
+        self.variables = []
+        self.units = []
+
+        self.port_manager = True
+        self.port_types = ["COM"]
+        self.port_properties = {
+            "EOL": "\n",
+            "timeout": 15,
```
afischer-sweepme, line 57: I guess a timeout of 5 s or so should work as well as the Arduino probably responds quite fast to each command and the initial communication during 'disconnect' typically takes about 2 s.

src/Switch-Arduino_MCP4728/main.py
```diff
@@ -0,0 +1,269 @@
...
+            if channel_num == 1:
+                n = self.pin
+            self.variables.append(f"Channel{n}")
+            self.units.append(self.unit[self.sweepmode])
+
+    """ here, semantic standard functions start that are called by SweepMe! during a measurement """
+
+    def connect(self):
+        # Set Name/Number of COM Port as key
+        instance_key = f"mcp4728_{self.port.port.port}"
```
afischer-sweepme: I would create instance_key as class variable self.instance_key to use it later again in disconnect without redefining it. The key itself I would replace with a combination of driver name and port name by using the GUI parameters

src/Switch-Arduino_MCP4728/main.py
```diff
@@ -0,0 +1,269 @@
...
+
+        if instance_key not in self.device_communication:
+            # Wait for Arduino initialization
+            self.port.read()
+            self.device_communication[instance_key] = "Connected"
+
+    def configure(self):
```
afischer-sweepme, line 143: The driver also needs an 'unconfigure' function. Otherwise, all channels stay at their last value. Here, I would set all channels back to 0.

afischer-sweepme (changes requested):
I think it is already a nice and clean driver. Adding a short pysweepme example to the folder "examples" could be helpful if you have already one.

src/Switch-Arduino_MCP4728/main.py
```diff
@@ -158,6 +158,17 @@ def configure(self):
             self.set_vref(use_internal_vref=False)
             self.max_voltage = self.external_voltage
 
+    def unconfigure(self):
+        # Set all outputs to 0
+        if self.multi_pins:
+            for address in self.addresses:
+                self.set_address(address)
+                for pin in range(4):
+                    print(pin)
```
afischer-sweepme, line 167: This print could be removed to keep the Debug widget empty if there are no problems.

afischer-sweepme (changes requested):
only a minor remaining point

afischer-sweepme (approved):

PR #30, opened 2024-02-02
Accretech uf contact wafer

afischer-sweepme:
The driver now support subsite stepping based on the A command. 
There is still an issue with SRQ events when multiple GPIB instruments are connected to the same GPIB adapter, but the driver is basically robust and works productively.

### DIFF
--- a/src/WaferProber-Accretech_UFseries/libs/accretech_uf.py
+++ b/src/WaferProber-Accretech_UFseries/libs/accretech_uf.py
@@ -63,7 +63,8 @@ class AccretechProber:
                 "No connection established with Accretech UF prober. "
                 "Please check port address / instrument is connected"
             )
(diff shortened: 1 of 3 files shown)


fk3:
> The driver now support subsite stepping based on the A command. There is still an issue with SRQ events when multiple GPIB instruments are connected to the same GPIB adapter, but the driver is basically robust and works productively.

What is the "A" command and where in the code was it added?

afischer-sweepme:
The A command is for absolute positioning via coordinates. It was already introduced in a prior PR. This PR just contains some new functions to register/unregister the SRQ event, Furthermore, the wafer is now contacted. Beforehand, the user had to enable this features manually in the code.

fk3 (approved):

PR #31, opened 2024-02-02
Switch ftdi ftd2xx

franz-sweepMe:


src/Switch-FTDI_FTD2XX/main.py
```diff
@@ -0,0 +1,143 @@
...
+
+import binascii
+
+import ftd2xx as ftd
+from pysweepme import EmptyDevice
+
+
+class Device(EmptyDevice):
+    description = """
+        Need FTDI driver installed
```
afischer-sweepme, line 51: Can you specify in which way a FTDI driver must be installed. Are there any dll files or libraries that must be installed additionally?

src/Switch-FTDI_FTD2XX/main.py
```diff
@@ -0,0 +1,143 @@
...
+# SOFTWARE.
+
+
+# SweepMe! device class
+# Device: Arduino MCP4728
+
+import os
+import sys
+
+file_dir = os.path.dirname(__file__)
```
afischer-sweepme, line 35: The entire sys.path.append block can be replaced
by 

from pysweepme import FolderManager
FolderManager.addFolderToPATH()

which is a convenience function to add the path of the libs folders. However, the directory of __file__ will not be added because it is not needed.

However, the code will not work for SweepMe! 1.5.5 because in this version addFolderToPATH added also subfolders of libs to sys.path which does not work for the ftd2xx library because it contain a subpackage with the same name.

Therefore, I suggest to omit support for SweepMe! 1.5.5. by removing the "libs" folder that is only there for 1.5.5 compatibility.

src/Switch-FTDI_FTD2XX/main.py
```diff
@@ -0,0 +1,143 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+
+# SweepMe! device class
+# Device: Arduino MCP4728
```
afischer-sweepme, line 30: device name FTDI xx245R or so

src/Switch-FTDI_FTD2XX/main.py
```diff
@@ -0,0 +1,143 @@
...
+class Device(EmptyDevice):
+    description = """
+        Need FTDI driver installed
+        Can send initial command when initializing or end command when deinitializing
+    """
+
+    def __init__(self) -> None:
+        EmptyDevice.__init__(self)
+
+        self.shortname = "FTDI245"
```
afischer-sweepme, line 58: if the driver fits to FTD2xx, we should change it here as well.

src/Switch-FTDI_FTD2XX/main.py
```diff
@@ -0,0 +1,143 @@
...
+
+        self.device.write(command)
+
+    def call(self):
+        # Number of bytes in queue
+        queue = self.device.getQueueStatus()
+
+        # check timeout?
+
+        if queue > 0:
```
afischer-sweepme, line 138: If you like to return a message string, it must be done in any case. The code works so far because the case where there are some characters in the buffer did not happen so far.
If a value is returned in 'call', self.variables, self.units, etc. must be adapted accordingly.

src/Switch-FTDI_FTD2XX/main.py
```diff
@@ -0,0 +1,143 @@
...
+        # TODO: Timeout?
+        pass
+
+    def apply(self):
+        # command: 5 Bytes
+
+        if self.encoding == "HEX":
+            # length 5
+            command = binascii.unhexlify(self.value.replace(" ", ""))
+        # elif self.encoding == "ASCII":
```
afischer-sweepme, line 125: remove if not needed

src/Switch-FTDI_FTD2XX/main.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def unconfigure(self):
+        # TODO: Timeout?
+        pass
+
+    def apply(self):
+        # command: 5 Bytes
+
+        if self.encoding == "HEX":
+            # length 5
```
afischer-sweepme, line 123: Can you explain length 5 or length 14 in more detail or remove it if not needed. Alternatively, one could add for each case a short example which kind of data is accepted.

src/Switch-FTDI_FTD2XX/main.py
```diff
@@ -0,0 +1,143 @@
...
+        self.instance_key = f"{self.driver_name}_{self.port_str}"
+
+        if self.instance_key not in self.device_communication:
+            self.device_communication[self.instance_key] = "Connected"
+
+    def disconnect(self):
+        # Close device
+        self.device.close()
+
+        if self.instance_key in self.device_communication:
```
afischer-sweepme, line 109: If self.device_communication is not used to make a decision in one of the driver instances, this part is not needed and can also be removed (as well as in 'connect'). 
Still, I suggest to check what happens if you have two modules in different branches that use the same driver.

src/Switch-FTDI_FTD2XX/main.py
```diff
@@ -0,0 +1,143 @@
...
+            self.device_communication.pop(self.instance_key)
+
+    def configure(self):
+        self.device.setTimeouts(5000, 5000)
```
afischer-sweepme, line 113: The read timeout could be a user parameter, but this can be also changed later when needed. One could define variables that are defined in __init__ to explain the meaning of the two values.

src/Switch-FTDI_FTD2XX/main.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def __init__(self) -> None:
+        EmptyDevice.__init__(self)
+
+        self.shortname = "FTDI245"
+        self.variables = []
+        self.units = []
+
+        self.port_manager = False
+        self.port_types = ["USB"]
```
afischer-sweepme, line 63: The port types are used by the port manager to automatically add ports to the list of ports after clicking the button "Find ports" without doing this in an "find_ports" method of the driver. The port type "USB" refers to "USBTMC" which is not related to the USB port used by the FTDI chip. So this line can be removed here.

afischer-sweepme (changes requested):
Thanks for creating the first version of the driver. I have some comments but there are no big changes.

src/Switch-FTDI_FTD2XX/main.py
franz-sweepMe: added the device instance as value in the device_communication dict to enable running 2 instances of the same driver

src/Switch-FTDI_FTD2XX/main.py
franz-sweepMe: that was a reference to the python package ftd2xx, which is included in the library folder

.gitignore
```diff
@@ -171,3 +171,5 @@ cython_debug/
 /src/*/libs/**/__pycache__/
 /src/*/libraries/libs_39_32/**/__pycache__/
 /src/*/libraries/libs_39_64/**/__pycache__/
+
+/src/Switch-FTDI_FTD245XX/thermusb_example
```
fk3, line 175: Is this an example that should go into the examples folder? Then it should be moved there and does not need to be ignored.

Otherwise it looks really strange like "Hey, we have an example, but we don't show it."
Also need to consider, what's excluded from the repository is not saved in case your local files get deleted.

src/Switch-FTDI_FTD2xx/main.py
```diff
@@ -48,94 +49,113 @@ class Device(EmptyDevice):
...
+    return [dev.decode() for dev in ftd.listDevices()]
+
+
+def set_GUIparameter(self):
+    return {
+        "Encoding": ["HEX", "ASCII"],
+        "SweepMode": "Command",
+        "Start command": "",
+        "End command": "",
+        "Readout length": 10,
```
afischer-sweepme, line 76: What about using a default readout length of 0 so that people need to switch it on when really needed.

afischer-sweepme (changes requested):

afischer-sweepme (approved):

afischer-sweepme (changes requested):

afischer-sweepme (approved):

PR #32, opened 2024-02-05
Mb switch revision

afischer-sweepme:
The driver for the mb Technologies switch is now ready to use and already productively used.

### DIFF
--- a/src/Switch-mbTechnologies_HVM/main.py
+++ b/src/Switch-mbTechnologies_HVM/main.py
@@ -59,6 +59,9 @@ class Device(EmptyDevice):
                     <li>Using a "0" all channels will remain open. This command is not part of the communication
                     protocol but was implemented in the SweepMe! driver for reasons of convenience.</li>
                     <li>TCPIP/LAN communication is possible by registering a raw socket with port 10001.</li>
+                    <li>In case of using the resistance box mode, the digital output pin 1 is set to High and a
+                     resistance of 300 kOhm is used. After the processing the branch, the digital output pin 1 is
+                      set to Low leading to a resistace of 0 Ohm (short)</li>
                     </ul>
                 """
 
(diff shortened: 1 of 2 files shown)


src/Switch-mbTechnologies_HVM/main.py
```diff
@@ -192,63 +210,75 @@ def get_identification(self):
...
+        return self.port.read() == "1"
 
     def reset(self) -> None:
-        """
-        Resets the instrument and opens all relays. 
+        """Resets the instrument and opens all relays.
         """
         self.port.write("*RST")
+        if not self.port_string.startswith("GPIB"):
+            self.port.read()
```
fk3, line 226: I've seen this line several times. Why is GPIB excluded? Is this something specific to this instrument? Or is that common knowledge for GPIB in general?

src/Switch-mbTechnologies_HVM/main.py
```diff
@@ -192,63 +210,75 @@ def get_identification(self):
...
+
         Response: 
         0 Self-test passed 
         n (relay, …) Self-test failed , n=number of failing relays, list of failing relays follows 
         """
         self.port.write("*TST?")
+        return self.port.read()
+
+    def run_self_test(self) -> str:
+        self.start_self_test()
```
fk3, line 248: The start_self_test() function has some important remarks in the docstring. If a user would use the run_self_test() function instead, he would not notice this information. Maybe it should also be mentioned here. And maybe a few more words on when a user should use start_self_test() vs run_self_test(). From the names of the functions they could mean the same.

src/Switch-mbTechnologies_HVM/main.py
afischer-sweepme: It is related to the software protocol of the instrument. In case of GPIB no answer is sent for function hat set a property, whereas for all other interfaces a message is always returned.

src/Switch-mbTechnologies_HVM/main.py
afischer-sweepme: I added further docu and renamed enquiry_self_test

src/Switch-mbTechnologies_HVM/main.py
afischer-sweepme: I added a further note to the code.

PR #43, opened 2024-03-13
Korad kd3005p

afischer-sweepme:
The driver now reads the instrument identifier omitting \x00 characters that make problems when printing them. To clean up some commands are wrapped into functions. and commented code is removed.
The driver is tested and works.

### DIFF
--- a/src/SMU-KORAD_KD3005P/main.py
+++ b/src/SMU-KORAD_KD3005P/main.py
@@ -156,10 +156,20 @@ class Device(EmptyDevice):
         return iset
 
     def set_output(self, state):
+        """
+        Sets the output state of the power supply to On or Off.
 
-        if not state:
+        Args:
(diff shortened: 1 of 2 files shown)


src/SMU-KORAD_KD3005P/main.py
```diff
@@ -150,7 +134,32 @@ def measure(self):
...
+        self.port.write("ISET1?")
+        iset = float(self.port.read(6)[:-1])
+        return iset
+
+    def set_output(self, state):
+
+        if not state:
+            state = 0
+        else:
+            state = 1
```
fk3, line 163: This function could use a docstring.
set_output can be misunderstood as applying a certain voltage to the output. Also what is state? Bool or int? Probably both would work, but that just adds confusion on how it is intended to use this function

fk3 (approved):

PR #44, opened 2024-03-27
Fixed minor errors for SMU-Agilent_B29xx

A-D-Fuchs:
Fixed the self.channel variable
Fixed the four-wire setting
Removed double space from apply string

### DIFF
--- a/src/SMU-Agilent_B29xx/main.py
+++ b/src/SMU-Agilent_B29xx/main.py
@@ -90,7 +90,9 @@ class Device(EmptyDevice):
         #     self.average = 100
             
         self.device = parameter['Device']
-        self.channel = self.device[-1]
+        # The channel is defined by the route_out parameter
+        # For 'Front' the channel is 1, for 'Rear' the channel is 2
+        self.channel = 1 if self.route_out == 'Front' else 2
         
         
     def initialize(self):
@@ -154,9 +156,9 @@ class Device(EmptyDevice):
         
         # 4-wire sense
         if self.four_wire:
-            self.port.write("SYST:REM ON")
+            self.port.write("SENS:REM ON")
         else:
-            self.port.write("SYST:REM OFF")
+            self.port.write("SENS:REM OFF")
         
         """
         # averaging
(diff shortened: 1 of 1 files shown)


src/SMU-Agilent_B29xx/main.py
```diff
@@ -90,7 +90,9 @@ def get_GUIparameter(self, parameter = {}):
         #     self.average = 100
             
         self.device = parameter['Device']
-        self.channel = self.device[-1]
+        # The channel is defined by the route_out parameter
+        # For 'Front' the channel is 1, for 'Rear' the channel is 2
+        self.channel = 1 if self.route_out == 'Front' else 2
```
afischer-sweepme, line 95: The channel is here related to models that have two SMU separate channels, i.e. dual-channel SMUs.
The driver is a bit old. At the beginning the channel information was derived from a suffix that was added to the driver name. New SweepMe! versions come with an SMU module that has a field with the key "Channel" which can be used to add possible channel options e.g. "1" and "2". The driver must be changed to the new standard.
I can do this if you can test the driver.
The route out option can then additionally be used to switch between the connectors at the front or at the rear.

afischer-sweepme (commented):
Thanks for creating a pull request.

src/SMU-Agilent_B29xx/main.py
afischer-sweepme: I think I understand now the intention of your change. 

The current version of this driver is not compatible with pysweepme. The multichannel variable is deprecated and must be replaced with using the field "Channel". I will create a new branch and remove the use of multichannel so that you can change your code to use the key "Channel" instead of "RouteOut".

A-D-Fuchs:
The changes in contrib/b29xx work for me. 
There is a format specifier missing in line 122. This does not allow the setting of currents. It should probably look like this:

```python
self.port.write(":SENS%s:VOLT:PROT %s" % (self.channel, self.protection))
```

afischer-sweepme:
Thanks for highlighting this. I added it and pushed a new version. Basically, the driver is missing functions that wraps each single command. Then, it would be also possible to write more individual scripts with pysweepme for the B29xx models
I will now merge the branch. If there is anything else, we can open a new one. Thanks for taking a look at the driver!

PR #45, opened 2024-03-27
LCRmeter Keithley 4200 SCS

franz-sweepMe:
Create LCRmeter driver based on SMU Keithley 4200 SCS driver.


afischer-sweepme:
I pushed some changes instead of commenting. I switched off GPIB/TCPIP communication via KXCI, card_id and card_name are defined now in __init__. The card_name is probably always "CVU1". The corresponding card_id will be automatically retrieved in connect. I removed the sweep mode for "Voltage level in V" to make the definition of the return variables easier.
The new pylptlib version must be copied to the driver as it is needed if SweepMe! is running directly on the 4200-SCS to load the lptlib.dll without the lptlib server. Whether range parameter is needed depends on whether there is command to change the range.

src/LCRmeter-Keithley_4200-SCS/main.py
```diff
@@ -0,0 +1,348 @@
...
+
+    def configure(self) -> None:
+        """Set bias and measurement parameters with start values from GUI."""
+        self.set_frequency(self.frequency)
+        self.set_delay(self.trigger_delay)
+
+        self.set_dc_bias(self.value_bias)
+        self.set_ac_voltage(self.value_level_rms)
+
+        self.set_measure_range()
```
afischer-sweepme, line 259: set_measure_range uses the self.range being "Auto". This handover of information via class variables is often a bit hidden and not easy to read. 
An alternative would be self.set_measure_range(self.range) and the function expects a string like "Auto" or enum like 0 (or both can be processed).

src/LCRmeter-Keithley_4200-SCS/main.py
```diff
@@ -0,0 +1,348 @@
...
+    def set_GUIparameter(self) -> dict:  # noqa: N802
+        """Set initial GUI parameter in SweepMe!."""
+        return {
+            "SweepMode": ["None", "Frequency in Hz", "Voltage bias in V"],
+            "StepMode": ["None", "Frequency in Hz", "Voltage bias in V"],
+            "ValueRMS": 0.02,
+            "ValueBias": 0.0,
+            "Frequency": 1000.0,
+            "Integration": list(self.speeds),
+            "ALC": ["Off"],  # Currently unused
```
afischer-sweepme, line 155: If ALC is not used at all it could be also commented out so that the the user interface for this field gets greyed out.

afischer-sweepme (changes requested):
The driver looks already really good!

afischer-sweepme (approved):
A very nice new driver :)

PR #47, opened 2024-03-28
Community standards

franz-sweepMe:
Add templates and code of conduct to fulfill GitHub Community standards for open source projects.

Open questions:
Please review the general license information


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
fk3, line 1: I think we can't provide a global license for the whole repository.
Each driver has it's own license.txt, and libraries also have their own license files (which are often not MIT).

If the community guidelines require a license file (and I would prefer to name it LICENSE.md), we can write an information that drivers and libraries each have their own license which can be found in the corresponding folder of the driver.

afischer-sweepme: That's a good point to add a LICENSE.md that makes once more clear that each driver comes with an individual license.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,128 @@
...
+Violating these terms may lead to a permanent ban.
+
+### 4. Permanent Ban
+
+**Community Impact**: Demonstrating a pattern of violation of community
+standards, including sustained inappropriate behavior,  harassment of an
+individual, or aggression toward or disparagement of classes of individuals.
+
+**Consequence**: A permanent ban from any sort of public interaction within
+the community.
```
fk3, line 113: I just checked how bans could be enforced on github. As it seems, there is a way to block users:
https://docs.github.com/en/communities/maintaining-your-safety-on-github/blocking-a-user-from-your-organization

However, this has a few more implications, like the reset of votes, stars, ... and the inability to fork, watch, pin our repositories. Maybe we should also mention that here or link the github article to be transparent on what a ban means technically?

And on another note: Should we specify a concrete time frame for temporary ban? like 1 week?

franz-sweepMe: This code of conduct is a template directly from GitHub and is used in that form in many projects. Therefore I would keep it in the original form.

PR #49, opened 2024-03-28
Keithley 2400 list sweep

afischer-sweepme:
List sweep feature added after contribution by user


src/SMU-Keithley_2400/main.py
```diff
@@ -106,6 +117,43 @@ def get_GUIparameter(self, parameter={}):
...
+        # use None (later AUTO), if empty
+        try:
+            self.listsweep_delay = float(parameter["ListSweepDelaytime"])
+        except:
+            self.listsweep_delay = None
+
+        # in case of a list sweep, the time stamp of each measurement point must be saved additionally
+        # to every (V,I) tuple
+        if self.sweepvalue == "List sweep":
+            self.variables = ["Voltage", "Current", "Timestamp"]
```
fk3, line 147: Does the device really return a timestamp? From the unit `s` I guess it's rather the (elapsed) time?

afischer-sweepme: It is right that this is rather some time since the start of the last instrument reset or since starting the instrument that indicates when the measurement was done and it is a plotable number. One could also call it just "Time", but "Timestamp" is also not completely incorrect as it is often the wording of the manufacturer. For example it is also named "relative timestamp" in the documentation.

src/SMU-Keithley_2400/main.py
```diff
@@ -25,72 +25,83 @@
...
+            "ListSweepDelaytime": 0.1,
         }
-                        
+
         return gui_parameter
-                                 
+
     def get_GUIparameter(self, parameter={}):
+
+        self.sweepvalue = parameter["SweepValue"]
```
fk3, line 103: a default for SweepValue should be added in set_GUIparameter, otherwise the driver fails in pysweepme with
`KeyError: 'SweepValue'`

afischer-sweepme: thanks, added it now

fk3 (approved):

PR #50, opened 2024-04-03
Temperature Accretech UFseries

franz-sweepMe:
Enable temperature control of Accretech UF devices.
The current version is more of a skeleton, as I do not know which additional configure and initialization steps are needed.

Open Questions:
- Is handling the port object using self.device_communication sufficient to enable parallel communication with a WaferProber and Temperature module?
- Which is the correct request: Chuck temperature or Hot chuck temperature?
- Are additional Temperature units besides °C needed?
- Are configure/initialization steps missing?
- reach needed?

### DIFF
--- a/src/Temperature-Accretech_UFseries/main.py
+++ b/src/Temperature-Accretech_UFseries/main.py
@@ -1,0 +1,136 @@
+# This Device Class is published under the terms of the MIT License.
+# Required Third Party Libraries, which are included in the Device Class
+# package for convenience purposes, may have a different license. You can
+# find those in the corresponding folders or contact the maintainer.
+#
+# MIT License
+#
+# Copyright (c) 2024 SweepMe! GmbH (sweep-me.net)
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# SweepMe! device class
+# Type: Temperature
+# Device: Accretech UF series
+
+
+from pysweepme.EmptyDeviceClass import EmptyDevice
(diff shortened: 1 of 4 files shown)


src/Temperature-Accretech_UFseries/main.py
```diff
@@ -0,0 +1,160 @@
...
+# standard import
+# import accretech_uf
+# importlib.reload(accretech_uf)
+
+# direct import by path
+import imp
+import os
+
+accretech_uf = imp.load_source(
+    "accretech_uf",
```
afischer-sweepme, line 47: To prevent some cross-talk between different versions of accretech_uf, it would make sense to name it here "accretech_uf_temperature". The variable in the line above does not need to be changed. It is just about the name that is used to register the module in the python system.

src/Temperature-Accretech_UFseries/main.py
```diff
@@ -0,0 +1,160 @@
...
+    def connect(self) -> None:
+        """Connect to the device and handle multiple instances."""
+        # creating an AccretechProber instance that handles all communication
+        # TODO: Handle multiple instances
+        self.instance_key = f"Accretech_UF_{self.port_str}"
+
+        if self.instance_key in self.device_communication:
+            self.port = self.device_communication[self.instance_key]
+        else:
+            self.device_communication[self.instance_key] = self.port
```
afischer-sweepme, line 110: This section could be commented out for now. The GPIB port itself is handled by SweepMe's port manager, so that each driver instance automatically gets the same GPIB port. 
However, it might be that we need the self.device_communication to mark whether the SRQ event is already registered or not.
If this can be done only once, we probably would need to also revise the accretech_uf library because it would need a switch to decide whether it should register for the SRQ event during connect. But this is something that we can check once the driver is tested.

src/Temperature-Accretech_UFseries/main.py
```diff
@@ -0,0 +1,160 @@
...
+        EmptyDevice.__init__(self)
+
+        self.instance_key = None
+        self.port_str = None
+        self.prober: accretech_uf.AccretechProber = None
+        self.temperature: float = 0.0
+
+        self.shortname = "Accretech"
+        self.variables = ["Temperature"]
+        self.units = [""]  # unit of X and Y is either µm or inch, depending on system settings
```
afischer-sweepme, line 74: the comment is still the one about the positional units

src/Temperature-Accretech_UFseries/main.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def apply(self) -> None:
+        """Apply the given temperature."""
+        temperature = self.value
+
+        max_temperature = 200.0
+        min_temperature = -55.0
+        if temperature > max_temperature or temperature < min_temperature:
+            msg = f"Accretech UF series: Temperature must be between {min_temperature} and {max_temperature} °C."
+            raise Exception(msg)
```
afischer-sweepme, line 139: I think this part should move into set_chuck_temperature in accretech_uf to have a general check there if needed.

I just see now that this check is already there, so you can remove it here.

src/Temperature-Accretech_UFseries/main.py
```diff
@@ -0,0 +1,160 @@
...
+
+        status = self.prober.set_chuck_temperature(temperature)
+        successful_status = 93
```
afischer-sweepme, line 142: Not fully sure yet, but probably it would be better to directly throw an error in accretech_uf.py if set_chuck_temperature fails. In this case the user needs to do some steps and the measurement should stop.

src/Temperature-Accretech_UFseries/main.py
```diff
@@ -0,0 +1,160 @@
...
+    def initialize(self) -> None:
+        """Initialize the device."""
+        # TODO: Check
+        self.prober.reset_alarm()
+
+    def configure(self):
+        pass
+
+    def unconfigure(self):
+        pass
```
afischer-sweepme, line 129: could be removed and added again when needed

afischer-sweepme (changes requested):
I have a few points. When changing the accretech_uf file, please copy it to the Accretech wafer prober driver as well to keep both files synced.

afischer-sweepme (approved):

PR #52, opened 2024-04-12
Logger PREVAC TM1x

franz-sweepMe:
Driver for the PREVAC Thickness Monitors TM13 and TM14.

The driver can request the frequency but fails after 10-30s of repeated requests to the device.

The functions for changing the device address and logic group are included but should only be used with caution, as a change in the device parameters can cause problems in communication with the TMC Controller unit.

### DIFF
--- a/src/Logger-PREVAC_TM1x/main.py
+++ b/src/Logger-PREVAC_TM1x/main.py
@@ -74,10 +74,10 @@ class Device(EmptyDevice):
         self.device_version: str = ""
         self.sample_rate: float = 1.0
         self.sample_rate_dict = {
-            "10": 1,
-            "4": 2,
-            "2": 3,
-            "1": 4,
+            "10.0": 1,
+            "4.0": 2,
+            "2.0": 3,
+            "1.0": 4,
             "0.5": 5,
         }
 
@@ -104,7 +104,7 @@ class Device(EmptyDevice):
     def set_GUIparameter(self) -> dict:  # noqa: N802
         """Get parameters from the GUI and set them as attributes."""
         return {
-            "Sample rate in Hz": list(self.sample_rate_dict),
+            "TM14 sample rate in Hz": list(self.sample_rate_dict),
             "Material density in g/cm^3": 1.0,
             "Impedance ratio": 1.0,
             "Tooling factor": 1.0,
@@ -114,7 +114,7 @@ class Device(EmptyDevice):
 
     def get_GUIparameter(self, parameter: dict) -> None:  # noqa: N802
         """Get parameters from the GUI and set them as attributes."""
-        self.sample_rate = float(parameter["Sample rate in Hz"])
+        self.sample_rate = float(parameter["TM14 sample rate in Hz"])
 
         self.material_density = float(parameter["Material density in g/cm^3"])
         self.impedance_ratio = float(parameter["Impedance ratio"])
@@ -126,12 +126,8 @@ class Device(EmptyDevice):
     def initialize(self) -> None:
         """Get initial frequency."""
         self.device_version = self.get_device_version()
-
-        # TM13 only supports a sample rate of 1 Hz
-        if self.device_version == 13:
-            self.sample_rate = 1.0
-
         self.reset_thickness()
+
         if self.initial_frequency == 0:
             msg = "PREVAC TM1x: Initial frequency is 0. Check the connection to crystal head."
             raise Exception(msg)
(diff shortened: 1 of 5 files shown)


src/Logger-PREVAC_TM1x/main.py
```diff
@@ -0,0 +1,370 @@
...
+        """Reset thickness by updating initial frequency."""
+        self.initial_frequency = self.get_frequency()
+
+    def calculate_rate(self, thickness: float) -> float:
+        """Calculate the rate of the thickness change in A/s."""
+        time_stamp = time.time()
+        if self.previous_thickness == 0 or self.previous_time == 0:
+            rate = 0
+        else:
+            rate = (thickness - self.previous_thickness) / (time_stamp - self.previous_time)
```
fk3, line 196: This could throw a division by zero error, if calculate rate is called more frequent than the system's time quantization.
In that case one could return the rate that was previously calculated (if available).

Same fallback could be applied if the numerator is zero and the function was called within a single sampling interval (if this is known).

src/Logger-PREVAC_TM1x/main.py
```diff
@@ -0,0 +1,379 @@
...
+        self.initial_frequency = self.get_frequency()
+
+    def calculate_rate(self, thickness: float) -> float:
+        """Calculate the rate of the thickness change in A/s."""
+        time_stamp = time.time()
+        if self.previous_thickness == 0 or self.previous_time == 0:
+            rate = 0
+        elif time_stamp == self.previous_time:
+            msg = "PREVAC TM1x: Time stamp is the same as the previous time stamp. calculate_rate is called too often."
+            raise Exception(msg)
```
fk3, line 198: Why not set `rate = self.previous_rate`, wouldn't this be more user friendly? Otherwise this branch just replaced one exception by another one

src/Logger-PREVAC_TM1x/main.py
```diff
@@ -0,0 +1,379 @@
...
+        elif thickness == self.previous_thickness:
```
fk3, line 199: This needs an additional condition like `and time_stamp - self.previous_time < sampling_interval`.
If the deposition really has stopped and the thickness does not change any more, it is normal that the thickness equals the previous thickness, and in this case the rate should really be zero, and not the previous_rate.

src/Logger-PREVAC_TM1x/main.py
```diff
@@ -0,0 +1,379 @@
...
+            # If calculate_rate is called faster than the sampling rate of the device, the thickness does not change.
+            # This would lead to a rate of 0. To avoid this, the rate is set to the previous rate.
+            rate = self.previous_rate
+        else:
+            rate = (thickness - self.previous_thickness) / (time_stamp - self.previous_time)
+            rate *= 10  # nm/s -> A/s
+
+        self.previous_thickness = thickness
+        self.previous_time = time_stamp
```
fk3, line 208: This would break the additional `and time_stamp - self.previous_time < sampling_interval` condition I've mentioned above. The previous_time must only be set if not returning the previous_rate. I.e. in the `self.previous_time == 0` case and the `else` case.

src/Logger-PREVAC_TM1x/main.py
```diff
@@ -0,0 +1,388 @@
...
+            rate = 0
+            self.previous_time = time_stamp
+
+        # Time stamp is the same as the previous time stamp, would give division by 0 error.
+        elif time_stamp == self.previous_time:  # noqa: SIM114
+            rate = self.previous_rate
+
+        # If calculate_rate is called faster than the sampling rate of the device, the thickness does not change.
+        # This would lead to a rate of 0. To avoid this, the rate is set to the previous rate.
+        elif thickness == self.previous_thickness and time_stamp - self.previous_time < 1 / self.sample_rate:
```
fk3, line 208: > `1 / self.sample_rate`
- From looking at `self.sample_rate_dict`, the variable `self.sample_rate` does not contain a frequency in Hz, but rather some index, or key? In that case, `1 / self.sample_rate` will produce unexpected results
- Not strictly an issue in Python 3, but in other languages or even Python 2, a division of two integers will always result in an integer, i.e. 1/4 = 0. To make the division yield a float, one of the arguments needs to be a float, e.g. by writing 1.0 / variable. As I said, it's not a problem in Python 3, as it will always give a float (e.g. 1/1 = 1.0), but I still prefer to make it explicit that a division of floats is intended.

fk3 (approved):

PR #55, opened 2024-04-12
TMC13-Shutter -> Initial Commit

franz-sweepMe:
A Switch driver to use the PREVAC TMC13 to open/close a connected Shutter.
The prevac_protocol is copied without modification from the Logger-PREVAC_TMC13 driver.

Open Questions:

- [x] Testing with Hardware
- [x] Is the shutter number needed in the set_shutter command?


afischer-sweepme:
The driver works now with a real shutter, I added actions for "close_shutter" and "open_shutter". Further I added a new variable "Open" that returns a bool to be able to plot it and use it for further evaluation.

fk3 (approved):

PR #57, opened 2024-04-23
image saving

sebb-oh:
save path like Camera-PC_Webcam,
file format functioning now,
changes units ms to s,
Bug Fixes for ROI parameters handover to camera


franz-sweepMe:
These are helpful additions to the driver, thank you for your contribution!

PR #63, opened 2024-05-03
Arduino_AllPins -> Handle variations in pin number of different models

franz-sweepMe:
Change handling of digital pins in .ino script.
Clean up/Commenting of main.py


src/Logger-Arduino_AllPins/main.py
```diff
@@ -69,30 +72,36 @@ def __init__(self) -> None:
...
             "EOL": "\n",
             "baudrate": 115200,
         }
+        self.port_str: str = ""
+        self.driver_name: str = ""
 
+        # Device Parameter
         self.max_voltage = 5.0
-
+        self.resolution: int = 10
```
fk3, line 88: What does a resolution of 10 mean? Is it 10 bit? In other places of the file self.resolution is rather the value range maximum, i.e. "resolution" is 1023 for 10 bit

src/Logger-Arduino_AllPins/Logger-Arduino_AllPins.ino
```diff
@@ -5,8 +5,7 @@ String EOL;
 
 void setup() {
 
-   for (int i=2; i <= 13; i++){
-  
+   for (int i=2; i <= NUM_DIGITAL_PINS; i++){
```
fk3, line 8: When starting with i=2 and ending with i=NUM_DIGITAL_PINS, in total one digital pin will be missed. Is this intended? And also, do all Arduinos start their digital pins on i=2?

src/Logger-Arduino_AllPins/main.py
franz-sweepMe: added a comment to clarify

src/Logger-Arduino_AllPins/main.py
```diff
@@ -69,30 +71,36 @@ def __init__(self) -> None:
...
             "EOL": "\n",
             "baudrate": 115200,
         }
+        self.port_str: str = ""
+        self.driver_name: str = ""
 
+        # Device Parameter
         self.max_voltage = 5.0
-
+        self.resolution: int = 4096  # number of steps for 10 bit resolution
```
fk3, line 87: 4096 would correspond to 12bit. Furthermore I think one needs to subtract 1, to be consistent with how self.resolution is calculated in get_GUIparameter

src/Logger-Arduino_AllPins/Logger-Arduino_AllPins.ino
```diff
@@ -5,17 +5,10 @@ String EOL;
 
 void setup() {
 
-   for (int i=2; i <= 13; i++){
-  
+   for (int i=0; i <= NUM_DIGITAL_PINS; i++){
     pinMode(i, INPUT);
    }
```
fk3, line 10: That's still inconsistent. Which are the pin numbers that need to be set to INPUT?

fk3: I mean, on one hand there is the channel, like D0, digital-channel zero. But it is not on Pin 0 of the arduino (or at least it would be a great coincidence)

fk3: Hm, forget the previous comment. I think arduino automatically maps the digital pin number to the correct physical port. But then I don't understand how the analog read can work with the driver, because it seems to work differently for the analog pins (which are referenced by name).

Another issue: NUM_DIGITAL_PINS includes the analog pins, which can be used in digital mode. In that case, all analog pins are configured as digital inputs and can no longer be used as analog ones?

franz-sweepMe: Analog pins can be used as general-purpose I/O pins as well. In case a pin has been set to output before, the pins must be set to input for the AnalogRead call to work. There is no extra command to set the Inpute-mode specifically to Analog Input.
https://docs.arduino.cc/learn/microcontrollers/analog-input/#details-and-caveats

fk3 (approved):

PR #65, opened 2024-05-21
MCC DAQ

franz-sweepMe:
raise ImportError for missing manufacturer installation in self.init to allow loading the driver in SweepMe GUI
Add Docstrings and Type Annotations, cleanup

### DIFF
--- a/src/Logger-MCC_DAQ/main.py
+++ b/src/Logger-MCC_DAQ/main.py
@@ -26,23 +26,24 @@
 
 # Contribution: We like to thank TU Dresden/Shayan Miri for providing the initial version of this driver.
 
-# SweepMe! device class
+# SweepMe! driver
 # *Module: Logger
-# *Device: Measurement Computing Corporation DAQ devices
+# *Instrument: Measurement Computing Corporation DAQ devices
 
 from pysweepme import addFolderToPATH
 from pysweepme.EmptyDeviceClass import EmptyDevice
 
 addFolderToPATH()
 
-# this driver needs libraries installed by the manufacturer software InstaCal
+# this driver needs libraries installed by the manufacturer software package Universal Library™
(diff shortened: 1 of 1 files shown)


src/Logger-MCC_DAQ/main.py
```diff
@@ -27,37 +27,43 @@
...
 
                 <p>If your device supports additional AI ranges, they can be added by extending the
                 <code>available_ai_ranges</code> dictionary.</p>
                 """
 
-    def __init__(self):
+    def __init__(self) -> None:
+        """Initialize driver parameters."""
         EmptyDevice.__init__(self)
+        if mcculw_library_missing:
```
afischer-sweepme, line 64: throwing an error in __init__ works but sometimes leads to the problem that the error message pops up to often, e.g. when the the driver is the first item in the device list of a module and one clicks on the combo box of the field "Device".

An alternative would be to add the check at the beginning of "find_ports" and at the beginning of "connect" so just before the library is really needed. Then, the user is immediately informed when searching for devices/ports or when a measurement is started

src/Logger-MCC_DAQ/main.py
```diff
@@ -27,37 +27,43 @@
 # Contribution: We like to thank TU Dresden/Shayan Miri for providing the initial version of this driver.
 
 # SweepMe! device class
-# Type: Logger
-# Device: Measurement Computing Corporation DAQ devices
+# *Module: Logger
+# *Device: Measurement Computing Corporation DAQ devices
```
fk3, line 31: This is the header from the Logger template:
# SweepMe! driver
# * Module: Logger
# * Instrument: Template
Maybe it's worth to adhere to those when changing the header anyway

afischer-sweepme (approved):

PR #66, opened 2024-05-21
MCP4728

franz-sweepMe:
Handle colons in self.value
Cleanup, add docstrings and type annotations

### DIFF
--- a/src/Switch-Arduino_MCP4728/main.py
+++ b/src/Switch-Arduino_MCP4728/main.py
@@ -25,9 +25,10 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+# SweepMe! driver
+# *Module: Switch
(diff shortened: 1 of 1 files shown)


src/Switch-Arduino_MCP4728/main.py
```diff
@@ -168,17 +206,20 @@ def unconfigure(self):
...
             self.set_voltage(self.pin, 0)
 
-    def apply(self):
+    def apply(self) -> None:
+        """Set the voltages to the pins."""
         if self.multi_pins:
             # Receive values as list, split by ":" and convert to float
             split_values = self.value.split(":")
+            # Replace , with . to enable float conversion
+            split_values = split_values.replace(",", ".")
```
afischer-sweepme, line 215: it must be added above the split. Otherwise it must be done elementwise.

Although I did it myself during my quick fix, could you implement it in a way that self.value is not overwritten. We had once a case where this lead to trouble, e.g. using

sweepvalue = self.value.replace(",", ".")
split_values = sweepvalue.split(":")

PR #67, opened 2024-05-23
MCC DAQ

franz-sweepMe:
Update the UL Library Import Error Handling to allow Driver Import in GUI
Update description with Pin assignment


afischer-sweepme (commented):
Should we maybe add 0 to the default to indicate counting starts from 0, e.g

"Analog input channels": "0, 1, 2",

PR #68, opened 2024-06-03
Accretech wafer prober driver

afischer-sweepme:
- Wafer prober adapted to support load_wafer and unload_wafer functions
- revised handling of sense_wafers at the start
- resetting subsite information when die has changed
- new functions added to accretech_uf library
- Temperature driver created


src/Temperature-Accretech_UFseries/libs/accretech_uf.py
```diff
@@ -184,7 +186,8 @@ def __init__(self, port: pysweepme.Ports.Port):
 
     def __del__(self) -> None:
         """When the object is deleted, the SRQ event is unregistered."""
-        self.unregister_srq_event()
+        pass
+        # self.unregister_srq_event()
```
fk3, line 190: Should better be deleted. It's not doing anything and the `__del__` function can be misleading. The separate register / unregister functions below are much better.

src/Temperature-Accretech_UFseries/libs/accretech_uf.py
```diff
@@ -275,7 +279,19 @@ def acquire_status_byte(self, timeout: float = 10.0) -> int:
                 print("-->", now.strftime("%H:%M:%S"), "Function:", function_calling_name)
 
         # Wait for the SRQ event to occur
-        response = self.port.port.wait_on_event(self.event_type, int(timeout * 1000))  # conversion from s to ms
+        starttime = time.time()
+        while True: 
+            if time.time()-starttime < timeout:
+                try:
+                    # response = self.port.port.wait_on_event(self.event_type, int(timeout * 1000))  # conversion from s to ms
```
fk3, line 286: Seems this line is no longer needed.
But I'm curious, why was it necessary to introduce a while loop?

src/Temperature-Accretech_UFseries/libs/accretech_uf.py
```diff
@@ -275,7 +279,19 @@ def acquire_status_byte(self, timeout: float = 10.0) -> int:
...
+                    response = self.port.port.wait_on_event(self.event_type, int(1000))  # waiting 1000 ms
+                    break
+                except pyvisa.errors.VisaIOError:
+                    # Timeout error
```
fk3, line 290: Is VisaIOError always a timeout? What happens if it is something different than a timeout error?

src/Temperature-Accretech_UFseries/main.py
```diff
@@ -113,26 +119,46 @@ def connect(self) -> None:
         #     self.device_communication[self.instance_key] = self.port
 
         self.prober = accretech_uf.AccretechProber(self.port)
-
         self.prober.set_verbose(self.verbosemode)
 
+        self.unique_identifier = "Driver_Accretech_UFseries_" + self.port_str
```
fk3, line 124: It's important that the Temperature and WaferProber driver use the same unique_identifier. Maybe it's worth adding a comment that mentions this fact. Just in case someone wants to change something in the future so they are directly informed to also change it for the other affected driver.

src/Temperature-Accretech_UFseries/main.py
```diff
@@ -113,26 +119,46 @@ def connect(self) -> None:
...
+        if not self.unique_identifier in self.device_communication:
+            self.device_communication[self.unique_identifier] = None
+
     def disconnect(self) -> None:
         """Disconnect from the device."""
-        del self.prober  # this makes sure the event mechanism is disabled
+        if self.unique_identifier in self.device_communication:
+            self.prober.unregister_srq_event()  # this makes sure the event mechanism is disabled
```
fk3, line 131: The unregister is called only once for a unique identifier. But in connect() above the register is called for every instance (in the `__init__`  of AccretechProber) even if the port is the same. This asymmetry does not make any problems?

src/Temperature-Accretech_UFseries/main.py
```diff
@@ -113,26 +119,46 @@ def connect(self) -> None:
...
+    def deinitialize(self):
+        pass
+
+    def unconfigure(self):
+        if self.idle_temperature != "":
+            _, target_temperature = self.prober.request_chuck_temperature()
+            if target_temperature != float(self.idle_temperature):
+                self.prober.set_chuck_temperature(float(self.idle_temperature))
+            else:
+                debug("Target temperature already set.")
```
fk3, line 148: "Target" temperature? Rather "Idle" temperature.
But I'm wondering if this message is really important or it should rather be shown in debugmode only.

src/Temperature-Accretech_UFseries/main.py
```diff
@@ -113,26 +119,46 @@ def connect(self) -> None:
...
+        
     def apply(self) -> None:
         """Apply the given temperature."""
-        self.prober.set_chuck_temperature(self.value)
+        _, target_temperature = self.prober.request_chuck_temperature()
+        if target_temperature != float(self.value):
+            self.prober.set_chuck_temperature(self.value)
+        else:
+            debug("Target temperature already set.")
```
fk3, line 156: Might as well be a debugmode only print?

src/WaferProber-Accretech_UFseries/libs/accretech_uf.py
fk3, line 1: Same comments as on the accretech_uf of the temperature controller. Remember to copy after addressing all comments above

src/WaferProber-Accretech_UFseries/main.py
```diff
@@ -5,7 +5,7 @@
 #
 # MIT License
 #
-# Copyright (c) 2022-2024 SweepMe! GmbH (sweep-me.net)
+# Copyright (c) 2022-2023 SweepMe! GmbH (sweep-me.net)
```
fk3, line 8: Hm, going back in time? :D

src/Temperature-Accretech_UFseries/main.py
```diff
@@ -87,19 +88,24 @@ def __init__(self) -> None:
...
             "TemperatureUnit": ["°C"],
+            "IdleTemperature": 30,
         }
 
     def get_GUIparameter(self, parameter: dict) -> None:  # noqa: N802
         """Handle SweepMe GUI parameters."""
+        self.port_str = parameter["Port"]
         self.units[0] = parameter["TemperatureUnit"]
         self.port_str = parameter["Port"]
+        self.idle_temperature = parameter["IdleTemperature"]
```
fk3, line 108: No need to set the port_str twice

src/WaferProber-Accretech_UFseries/main.py
```diff
@@ -172,15 +183,63 @@ def get_current_subsite(self):
...
+        if not self.unique_identifier in self.device_communication:
+            self.device_communication[self.unique_identifier] = None
+
     def disconnect(self):
-        del self.prober  # this makes sure the event mechanism is disabled
+
+        if self.unique_identifier in self.device_communication:
+            self.prober.unregister_srq_event()  # this makes sure the event mechanism is disabled
+            del self.device_communication[self.unique_identifier]
+        del self.prober
```
fk3, line 242: Same comments as for the temperature driver

src/WaferProber-Accretech_UFseries/main.py
```diff
@@ -196,25 +255,17 @@ def initialize(self):
...
+            
+            if self.prober.is_wafer_on_chuck():
+                self.prober.unload_all_wafers()
 
-            cassette_status = self.prober.request_cassette_status()
-            if cassette_status == "000000":
-                self.prober.sense_wafers()  # Wafer sensing "jw" command
+            self.check_and_sense_wafers()
 
+            """
```
fk3, line 268: Is this code obsolete? Then it can be removed.
Or is it needed when users have some special requirements and need to re-enable it? Then there should be an explanation when it is needed and how it should be used

src/WaferProber-Accretech_UFseries/main.py
```diff
@@ -261,140 +313,212 @@ def unconfigure(self):
...
+                    message_box("Lot process will be finished which takes 1-2 minutes.\n\n"
+                                "Do not terminate the run and please wait until the program finishes."
+                                "You can close this message when the run has completed normally.", blocking=False)
+
+                self.prober.preload_specified_wafer(9, 99)
+
+    def signout(self):
+        pass
 
     def apply(self):
```
fk3, line 332: The apply() function is huge and might benefit from a docstring that summarized what is happening in the function.
The comments within the code are already helpful, but don't give a bigger picture that might help when looking at the driver in half a year.

src/Temperature-Accretech_UFseries/libs/accretech_uf.py
afischer-sweepme: The while-loop is needed to be able to stop the waiting for the event in future. It is not yet implemented, but a Stop in the GUI of SweepMe! could also stop the waiting. So far one needs to wait until the end and some processes have a timeout of 300 s to make sure all wafers are loaded.

afischer-sweepme: added a comment

src/Temperature-Accretech_UFseries/main.py
afischer-sweepme: There was no problem because registering again does not registering it twice, but removing it leads to the fact that it cannot be removed again.
I will not change it as this would have to be tested otherwise.

src/Temperature-Accretech_UFseries/main.py
afischer-sweepme: Changed it to "Idle". The debug output can be removed if there are no problems in future. So far, it is always better if the Accretech driver is a bit more verbose to understand what is going on in case of a problem.

src/Temperature-Accretech_UFseries/main.py
afischer-sweepme: see above

src/WaferProber-Accretech_UFseries/main.py
afischer-sweepme: same answer as for the temperature driver

src/WaferProber-Accretech_UFseries/main.py
afischer-sweepme: I removed it

src/WaferProber-Accretech_UFseries/main.py
afischer-sweepme: added a doctstring

src/WaferProber-Accretech_UFseries/libs/accretech_uf.py
afischer-sweepme: done

fk3 (approved):

PR #70, opened 2024-06-07
Agilent b29xx fix 

afischer-sweepme:
Commands in configure have been revised
general cleanup


src/SMU-Agilent_B29xx/main.py
```diff
@@ -153,38 +160,39 @@ def configure(self):
             self.port.write(":SENSe:AVER:COUN 1")  
         """
 
-        self.port.write(":OUTP%s:PROT ON" % self.channel)
```
fk3, line 156: What was the purpose of OUTP:PROT ON? Enable output protection? Isn't it dangerous if that is removed?

fk3 (approved):

PR #75, opened 2024-06-27
Fixed depricated time.clock call

A-D-Fuchs:
Replaced the time.clock() call with time.perf_counter().

time.clock() was deprecated in Python 3.3. and removed in Python 3.8

This fix would not work on Python Versions < 3.3.

One could try and except that if you still wanted to support < 3.3


fk3 (approved):
Thanks for the fix!

PR #77, opened 2024-06-27
K2000 Close port on disconnect

A-D-Fuchs:
Added closing of the port on the disconnect function.

The port was never closed when we used this device, causing repeated script executions to fail. 
It seems the port was never closed at the end of using it in one script. 
Is this the only way or is there some better way to handle closing of ports at the end of a script execution with these drivers?

### DIFF
--- a/src/Logger-Keithley_2000/main.py
+++ b/src/Logger-Keithley_2000/main.py
@@ -241,7 +241,10 @@ class Device(EmptyDevice):
         answer = self.port.read()  # here we read the response from the "READ?" request in 'measure'
         #print("Response to READ? command:", answer)
 
-        return [float(answer)] 
+        return [float(answer)]
+
+    def disconnect(self):
+        self.port.close() 
         
         
     """


afischer-sweepme:
@fk3 Could it be related to the fact that pysweepme opens the port, but there is no mechanism to close the port in standalone python scripts. Normally, the port is "automatically" closed when the script ends, but depending on the development environment, the port might not be closed. In this case a re-execution of the script fails. Maybe a good reason to move the port opening/closing into the EmptyDevice class in pysweepme and insert a mechanisum to call something like Device.disconnect() instead of directly closing the port object.

@A-D-Fuchs Could you add whether you are using the driver with SweepMe! or with pysweepme for standalone python scripts?

A-D-Fuchs:
We are using pysweepme version 1.5.6.10 in our standalone scripts.

It is connected via RS232-Usb adapter. It seems to work much worse with the RS232 connection than with the GPIB-USB connection. With the GPIB connection it does not need to be manually disconnected and the reading speed with the RS232 is exactly 2s (this seems to be the timeout time, could there be some error occurring when reading/writing to RS232-USB adapters)?this seems to be the case for two different RS232-USB adapters, so I don't believe it is the adapters fault?

fk3:
> We are using pysweepme version 1.5.6.10 in our standalone scripts.

alright, thanks. I think then the issue is what Axel already pointed out. In SweepMe! after the measurement the port manager is instructed to close all opened ports, but this functionality is missing in standalone pysweepme.

In my opinion we should add this feature to pysweepme, to keep it symmetrical. If pysweepme opens the port, it should also close the port. And only if the driver opens the port (i.e. not using the port manager), the driver should also close the port.

I am going to add an item to our todo list.

But right now, the best way to close the port is to add this to your standalone script. I assume you are obtaining the driver instance with
```python
my_driver = get_driver(...)
```
right?

If you have a "simple" setup where each instrument is only addressed by a single driver and all drivers are using the port manager, you can simply use a
```python
my_driver.port.close()
```

Or for a more robust approach
```python
from pysweepme import get_driver, close_port

my_driver = get_driver(...)

# your actual measurement routines here

if my_driver.port_manager:
    close_port(my_driver.get_port())
```
(haven't tested it, though)

afischer-sweepme:
> It is connected via RS232-Usb adapter. It seems to work much worse with the RS232 connection than with the GPIB-USB connection. With the GPIB connection it does not need to be manually disconnected and the reading speed with the RS232 is exactly 2s (this seems to be the timeout time, could there be some error occurring when reading/writing to RS232-USB adapters)?this seems to be the case for two different RS232-USB adapters, so I don't believe it is the adapters fault?

Sounds like the typical timeout problem. Please check whether your Keithley 2000 uses the same terminator character as defined in the driver in self.port_properties being CR \r

https://github.com/SweepMe/instrument-drivers/blob/3ed40e392f0e2b1d1d80d80f05e6626466ad4304/src/Logger-Keithley_2000/main.py#L59

A-D-Fuchs:
> Sounds like the typical timeout problem. Please check whether your Keithley 2000 uses the same terminator character as defined in the driver in self.port_properties being CR \r
> 
> https://github.com/SweepMe/instrument-drivers/blob/3ed40e392f0e2b1d1d80d80f05e6626466ad4304/src/Logger-Keithley_2000/main.py#L59

That fixed it :)

> Or for a more robust approach
> 
> ```python
> from pysweepme import get_driver, close_port
> 
> my_driver = get_driver(...)
> 
> # your actual measurement routines here
> 
> if my_driver.port_manager:
>     close_port(my_driver.get_port())
> ```
> 
> (haven't tested it, though)

We now use `pysweepme.close_port(self.driver.get_port())` in our driver wrappers and it works nicely

fk3:
I'm closing this PR, as closing the port is handled outside of the driver.

PR #79, opened 2024-06-30
Dobot MG400 -> Jump mode added

afischer-sweepme:
Jump mode added for Dobot MG400.
The robot goes back to a given movement height when x, y or r coordinate change.

### DIFF
--- a/src/Robot-Dobot_MG400/main.py
+++ b/src/Robot-Dobot_MG400/main.py
@@ -5,7 +5,7 @@
 #
 # MIT License
 # 
-# Copyright (c) 2022 SweepMe! GmbH (sweep-me.net)
+# Copyright (c) 2022-2024 SweepMe! GmbH (sweep-me.net)
 # 
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -26,9 +26,9 @@
 # SOFTWARE.
 
 
-# SweepMe! device class
-# Type: Robot
-# Device: Dobot MG400
+# SweepMe! driver
+# * Module: Robot
+# * Instrument: Dobot MG400
 
 from FolderManager import addFolderToPATH
 addFolderToPATH()
@@ -40,9 +40,9 @@ import dobot_api
 import time
 import select
 
-from ErrorMessage import error, debug
+from pysweepme.ErrorMessage import error, debug
 
-from EmptyDeviceClass import EmptyDevice
+from pysweepme.EmptyDeviceClass import EmptyDevice
 
 
 class Device(EmptyDevice):
@@ -125,6 +125,11 @@ class Device(EmptyDevice):
                         "Payload x offset": 0.0,
                         "Payload y offset": 0.0,
                         # "Payload z offset": 0.0,
+                        " ": None,
+                        "Jump": None,
+                        "Use jump mode": False,
+                        "Movement height": 0.0,
+                        "  ": None,
 
                         "GoHomeStart": True,
                         "GoHomeEnd": True,
@@ -151,6 +156,9 @@ class Device(EmptyDevice):
         self.acceleration_factor = parameter["Acceleration factor"]
         self.global_speed_factor = parameter["Global speed factor"]
         self.speed_factor = parameter["Speed factor"]
+
+        self.use_jump_mode = parameter["Use jump mode"]
+        self.movement_height = parameter["Movement height"]
         
         self.variables = ["x", "y", "z", "r"] 
         self.units = [self.length_unit] * 3 + ["°"]
@@ -195,10 +203,21 @@ class Device(EmptyDevice):
 
     def configure(self):
 
+        self.movement_height = float(self.movement_height)
         self.set_acceleration_linear(self.acceleration_factor)  # Linear acceleration factor 1-100
         self.set_speed_global(self.global_speed_factor)  # Global speed factor 1-100
         self.set_speed_linear(self.speed_factor)  # Linear speed factor 1-100
-        self._last_xyzr = (None, None, None, None)
+        # self._last_xyzr = (None, None, None, None)
+        self._last_xyzr = self.get_position()
+
+    def unconfigure(self):
+
+        if self.use_jump_mode:
+            self.move_linear(self._last_xyzr[0], self._last_xyzr[1], self.movement_height, self._last_xyzr[3])
+            self.sync(self.reach_position_timeout)
+
+        mode = self.get_robot_mode()
+        # todo: check for errors
 
     def reconfigure(self, parameters, keys):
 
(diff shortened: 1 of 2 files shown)


src/Robot-Dobot_MG400/main.py
```diff
@@ -125,6 +125,11 @@ def set_GUIparameter(self):
                         "Payload x offset": 0.0,
                         "Payload y offset": 0.0,
                         # "Payload z offset": 0.0,
+                        " ": None,
+                        "Jump": None,
+                        "Use jump mode": False,
+                        "Movement height": 0.0,
```
fk3, line 131: maybe explain new parameters in description

Also there is a class variable "axes" which seems to be used nowhere?

src/Robot-Dobot_MG400/main.py
```diff
@@ -125,6 +125,11 @@ def set_GUIparameter(self):
...
+                        "  ": None,
```
fk3, line 132: This is a separator, right? And the different amount of spaces is just to have unique dictionary keys?
I think this is very unintuitive, can't we use a more describing dictionary key, maybe with an underscore to show it's not a real parameter, e.g.
```python
{
    ...
    "_separator1": None,
    ...
    "_separator2": None,
    ...
}
```

Also it appears a bit fragile, because it depends on the order of entries in a (unordered) dictionary. But that's another story.

src/Robot-Dobot_MG400/main.py
```diff
@@ -195,10 +203,21 @@ def deinitialize(self):
 
     def configure(self):
 
+        self.movement_height = float(self.movement_height)
```
fk3, line 206: I think the conversion to float should already happen in set_guiparameter, so that self.movement_height always has the right type, even if in the future someone wants to use it already in initialize etc

src/Robot-Dobot_MG400/main.py
```diff
@@ -195,10 +203,21 @@ def deinitialize(self):
...
         self.set_acceleration_linear(self.acceleration_factor)  # Linear acceleration factor 1-100
         self.set_speed_global(self.global_speed_factor)  # Global speed factor 1-100
         self.set_speed_linear(self.speed_factor)  # Linear speed factor 1-100
-        self._last_xyzr = (None, None, None, None)
+        # self._last_xyzr = (None, None, None, None)
```
fk3, line 210: This line can be cleaned up (together with a couple of old prints)

Alternatively, it could also be placed in init to make sure _last_xyzr is set at all

src/Robot-Dobot_MG400/main.py
```diff
@@ -195,10 +203,21 @@ def deinitialize(self):
...
+        self._last_xyzr = self.get_position()
```
fk3, line 211: I assume the purpose of this line is to ensure the robot does not move up and down if it already is at the correct position when the measurement starts?
Also assuming get_position is returning floats and the dobot is actually measuring the position (and not just returning the most recent set values), there will probably always be some deviation in the 0.001 range. Not necessarily an issue, just something to be aware of.

src/Robot-Dobot_MG400/main.py
```diff
@@ -241,15 +255,29 @@ def apply(self):
...
-            self._last_xyzr = (x, y, z, r)
+
+            if self.use_jump_mode:
+
+                # we only move to movement height if x, y, or r change
+                # in case of a z scan, we do not need to go back to movement height
+                if self._last_xyzr[0, 2] != (x, y) or self._last_xyzr[3] != r:
+
+                    # vertical move to movement height
+                    self.move_linear(self._last_xyzr[0], self._last_xyzr[1], self.movement_height, self._last_xyzr[3])
```
fk3, line 266: Something that probably depends on the application, but what happens (or what should happen) when the movement_hight is already lower than the current position? Probably no need to change the code for now, but might be worth mentioning in the description.

src/Robot-Dobot_MG400/main.py
```diff
@@ -195,10 +203,21 @@ def deinitialize(self):
...
         self.set_speed_linear(self.speed_factor)  # Linear speed factor 1-100
-        self._last_xyzr = (None, None, None, None)
+        # self._last_xyzr = (None, None, None, None)
+        self._last_xyzr = self.get_position()
+
+    def unconfigure(self):
+
+        if self.use_jump_mode:
+            self.move_linear(self._last_xyzr[0], self._last_xyzr[1], self.movement_height, self._last_xyzr[3])
+            self.sync(self.reach_position_timeout)
```
fk3, line 217: This could go into a `move_up()` function, seems the same lines are used in apply() as well. For symmetry reasons a `move_down()` function could be provided as well. It's not needed in the driver, but could be used in actions in the future?

src/Robot-Dobot_MG400/main.py
```diff
@@ -195,10 +203,21 @@ def deinitialize(self):
...
+
+        mode = self.get_robot_mode()
+        # todo: check for errors
```
fk3, line 220: Should the error checking be implemented before going to test the driver?

And secondly, checking for errors when the measurement is already over might be a bit late.

src/Robot-Dobot_MG400/main.py
afischer-sweepme: yes, it is separator. However, if you give use a non-empty key it will be used a bold label to indicate a section:
image
You are right that this is not a good solution, but it cannot be solved within this PR, but needs to be handled in SweepMe! by revising the handling of gui parameters.

src/Robot-Dobot_MG400/main.py
afischer-sweepme: addded further description and comment

src/Robot-Dobot_MG400/main.py
afischer-sweepme: Moved it to get_GUIparameter

src/Robot-Dobot_MG400/main.py
afischer-sweepme: When using the new jump mode the robot makes at the beginning a move to the movement height. Because move_linear always needs x, y, z, and r, it is necessary to read them out once at the beginning to be able to set the current x, y, and r when going to the movement height.

src/Robot-Dobot_MG400/main.py
afischer-sweepme: It would be something like `move_to_movement_height()`. Whether this is a move up or a move down depends on the start height of the move as you already mentioned in your other comment.

I will add no function for now, but after more testing, it might be worth to add it. A good action would be "go_home" which I just added.

src/Robot-Dobot_MG400/main.py
afischer-sweepme: Basically both can work, but because of the geometry of the robot samples are typically approached from the top.
At the end the code is very generic. It will move from all positions to the movement height. This also makes sense. In case the robot starts (e.g. because of an accident) at the same height like the samples, it would go first back to movement height before approaching a new sample.
I will not add a description because each user needs to test this anyway for each individual setup.

src/Robot-Dobot_MG400/main.py
afischer-sweepme: This line is already part of the productive driver version. `self.get_robot_mode` is not needed at the end of the driver. I think I used it once to check whether the robot is still responsive when there have been errors in configure during enabling the robot. In this case, one could have done some checks to present an error message.
Because the code it not used, I removed it

fk3 (dismissed):

fk3 (approved):

PR #80, opened 2024-07-03
LCRmeter Metrohm Autolab

franz-sweepMe:
Add a driver for Metrohm Autolab Potentiostat.


src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
```diff
@@ -0,0 +1,333 @@
...
+    <ul>
+    <li>Nova must be closed when starting the measurement.</li>
+    </ul>
+    """
+
+    def __init__(self) -> None:
+        """Initializes the device class."""
+        super().__init__()
+
+        self.variables = ["|Z|", "Phase", "Frequency", "Voltage bias"]
```
afischer-sweepme, line 84: Check whether |Z| is called "Magnitude" in other LCRmeter drivers

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
```diff
@@ -0,0 +1,333 @@
...
+        self.units = ["", "", "Hz", "V"]
```
afischer-sweepme, line 85: Unit of Magnitude is Ohm, unit of Phase is °

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
```diff
@@ -0,0 +1,333 @@
...
+        }
+
+    def find_ports(self) -> list[str]:
+        """Get available hardware setup files defined in the .ini file."""
+        try:
+            ini_data = self.get_configoptions("Autolab HardwareSetup")
+            hardware_setup_files = ini_data["hardware_setup_files"].replace("\n", "").split(",")
+        except KeyError:
+            hardware_setup_files = []
+            debug("No Autolab HardwareSetup files found in .ini file.")
```
afischer-sweepme, line 164: Maybe return some help how to setup the ini file to get started?

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
afischer-sweepme: Or is this rather R and X instead of |Z| and Phase. The LCRmeter module expected R-X to automatically post-process other values. The returned data seems to be R and X.

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
```diff
@@ -0,0 +1,333 @@
...
+            self.handle_set_value(self.sweepmode, sweep_value)
+
+        if self.stepmode != "None":
+            step_value = float(self.stepvalue)
+            self.handle_set_value(self.stepmode, step_value)
+
+    def measure(self) -> None:
+        """Retrieve Impedance results from device."""
+        # Run the measurement
+        self.Fra.Start()
```
afischer-sweepme, line 251: Is it really needed to use Start at each measurement point. Would it be possible to call Start once in 'configure'?

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
```diff
@@ -0,0 +1,333 @@
...
+        self.sweepmode = None
+        self.stepmode = None
+
+        # Measurement Parameters
+        self.frequency: float = 0.0
+        self.ac_amplitude: float = 0.0
+        self.dc_bias: float = 0.0
+        self.measure_range: str = ""
+        self.measurement_ranges = {
+            "1000A": "CR00_1000A",
```
afischer-sweepme, line 108: Space between number and unit? (e.g. compare with SMU drivers)

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
```diff
@@ -0,0 +1,333 @@
...
+class Device(EmptyDevice):
+    """Device class for the Metrohm Autolab PGSTAT used as LCRmeter."""
+
+    description = """
+    <h3>Metrohm Autolab PGSTAT</h3>
+    <p>This driver controls Metrohm Autolab Potentiostats.</p>
+    <h4>Setup</h4>
+    <ul>
+    <li>Install the Autolab SDK 2.1.</li>
+    <li>Create a device configuration of your pump using QmixElements of the CETONI SDK.</li>
```
afischer-sweepme, line 66: bullet point from another driver

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
```diff
@@ -0,0 +1,333 @@
...
+    from EcoChemie.Autolab.Sdk import Instrument
+    from EcoChemie.Autolab.SDK.Extensions import InstrumentExtensions
+
+except:
+    # TODO: Handle import error
+    import_failed = True
+
+
+class Device(EmptyDevice):
+    """Device class for the Metrohm Autolab PGSTAT used as LCRmeter."""
```
afischer-sweepme, line 58: Driver instead of Device class

src/LCRmeter-Metrohm_AutolabPGSTAT/tests/integration/tests.py
```diff
@@ -0,0 +1,37 @@
+import pytest
+import pysweepme
+
+
+driver_name = "LCRmeter-Metrohm_AutolabPGTSTAT"
+driver_path = r"/src/LCRmeter-Metrohm_AutolabPGSTAT"
+port_string = "COM5"
```
afischer-sweepme, line 7: Is this correct?

afischer-sweepme (commented):
Looks already very nice!

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
```diff
@@ -0,0 +1,333 @@
...
+    def configure(self) -> None:
+        """Set bias and measurement parameters with start values from GUI."""
+        # Initial Measurement Parameters
+        self.set_frequency(self.frequency)
+        self.set_dc_bias(self.dc_bias)
+        self.set_ac_voltage(self.ac_amplitude)
+
+        # General Measurement Parameters
+        self.Fra.WaveType = self.Fra.WaveType.Sine
+        self.Fra.MinimumIntegrationTime = 0.125
```
afischer-sweepme, line 220: Might be a parameter that the user needs to change. The LCRmeter module has a field "Integration" where one could provide typical values like 0.125 s, 0.25 s, 0.5 s, 1.0 s, 1.25 s, 2.5 s and so on. Or 0.1 s, 0.2 s, 0.5 s, 1.0 s, 2.0 s, 5.0 s ...

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
franz-sweepMe: The command is called Start, but acts more like a "measure" command. Without it, the values are not updated

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
franz-sweepMe: added a comment that this parameter can be added to the GUI if needed

src/LCRmeter-Metrohm_AutolabPGSTAT/LCRmeter-Metrohm_AutolabPGSTAT.ini
```diff
@@ -0,0 +1,5 @@
+[Autolab HardwareSetup]
+adk_path = C:\Program Files\Metrohm Autolab\Autolab SDK 2.1\Hardware Setup Files\Adk.x
+hardware_setup_files =
+    C:\Program Files\Metrohm Autolab\Autolab SDK 2.1\Hardware Setup Files\PGSTAT302N\HardwareSetup.AUT83940.xml,
```
fk3, line 4: Should this go into a published driver if the hardware configuration file is specific to a certain customer?

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
```diff
@@ -0,0 +1,333 @@
...
+
+import_failed = False
+try:
+    clr.AddReference(autolab_sdk_path)
+    clr.AddReference(extension_path)
+    from EcoChemie.Autolab.Sdk import Instrument
+    from EcoChemie.Autolab.SDK.Extensions import InstrumentExtensions
+
+except:
+    # TODO: Handle import error
```
fk3, line 53: As long as there is no specific error handling, I would recommend to print a debug message and re-raise the exception with `raise`. There is no need to load the driver if the SDK cannot be imported

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
fk3:
```suggestion
    """Driver for the Metrohm Autolab PGSTAT used as LCRmeter."""
```

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
```diff
@@ -0,0 +1,333 @@
...
+
+class Device(EmptyDevice):
+    """Device class for the Metrohm Autolab PGSTAT used as LCRmeter."""
+
+    description = """
+    <h3>Metrohm Autolab PGSTAT</h3>
+    <p>This driver controls Metrohm Autolab Potentiostats.</p>
+    <h4>Setup</h4>
+    <ul>
+    <li>Install the Autolab SDK 2.1.</li>
```
fk3, line 65: should there be a note about having to use a beta version because the officially latest version does not work?

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
```diff
@@ -0,0 +1,333 @@
...
+    <li>Create a device configuration of your pump using QmixElements of the CETONI SDK.</li>
+    <li>Copy the LCRmeter_Metrohm-AutolabPGSTAT.ini file to public documents/SweepMe!/CustomFiles. In this file:
```
fk3, line 66: Isn't there a convenience function for this by right-clicking on the version in the version manager or something like that?

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
```diff
@@ -0,0 +1,339 @@
...
+
+    def disconnect(self) -> None:
+        """Disconnect from the Metrohm Autolab LCRmeter."""
+        self.autolab.Disconnect()
+
+    def initialize(self) -> None:
+        """Initialize the Metrohm Autolab LCRmeter."""
+
+    def deinitialize(self) -> None:
+        """Reset device and close connection."""
```
fk3, line 213: Is it planned to include (de)initialization steps in the future?

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
```diff
@@ -0,0 +1,339 @@
...
+            step_value = float(self.stepvalue)
+            self.handle_set_value(self.stepmode, step_value)
+
+    def measure(self) -> None:
+        """Retrieve Impedance results from device."""
+        # Run the measurement
+        self.Fra.Start()
+
+        # Retrieve measured values
+        self.resistance, self.reactance = self.measure_resistance_reactance()
```
fk3, line 260: would it make sense to move this to request_result() or read_result() semantic functions?
Probably only makes a difference if self.Fra.Start() is non-blocking

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
```diff
@@ -0,0 +1,339 @@
...
+        """Read the ini file and set the adx path."""
+        try:
+            ini_data = self.get_configoptions("Autolab HardwareSetup")
+            adk_path = ini_data["adk_path"]
+        except KeyError:
+            adk_path = ""
+            debug("No adk_path found in ini file.")
+
+        if Path(adk_path).exists() is False:
+            msg = "Invalid adk_path. Ensure the LCRmeter_Metrohm-AutolabPGSTAT.ini file is set up correctly."
```
fk3, line 283: So if there is no adk_path I first get the debug message that adk_path is not found in ini file, and afterwards I get the exception that the adk_path in invalid. Not really a big issue, just wanted to point it out.

src/LCRmeter-Metrohm_AutolabPGSTAT/LCRmeter-Metrohm_AutolabPGSTAT.ini
fk3: I'm not complaining about the template ini itself, just the chosen xml file - because this particular xml file is not part of the SDK installation.

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
franz-sweepMe: The beta version is needed for older devices. in general it should work with the standard available SDK

src/LCRmeter-Metrohm_AutolabPGSTAT/main.py
franz-sweepMe: It should be possible to load the driver without raising an exception, otherwise it will not show up in the list of available drivers in the LCRmeter module. Then, a user would not know why the driver is not shown.

Additionally, I realized that in the future there might be different SDK Versions from 2.1 or they might be saved in a different location. To enable setting the SDK path in the .ini aswell, I moved the load_sdk function insde the driver

fk3 (approved):

PR #84, opened 2024-07-17
Update main.py

sebb-oh:
- Fixing image saving: FileFormat was just changing the file suffix, not the actual image format, resulting in naming JPEGs .png

- Added all supported file formats to the selection


fk3 (approved):
Thanks for the PR!

PR #86, opened 2024-08-06
multiple-pulse capability enabled

Code-Craftsman-Christian:
The capability to emit multiple pulses per sequence step was added to the driver. For this, the sweep mode was misused by setting start and stop voltage/current to the same value. 

As this also works for a single pulse as well as the DC mode, the initialization was modified to have the B290xB SMU run in the sweep mode all the time, even if the pulse count in the GUI is set to 1 or if the SMU is used for DC output. No drawbacks were observed in practical use other than that the symbol on the display of the SMU shows a sweep rather than a DC or pulse graphic.

### DIFF
--- a/src/SMU-Agilent_B29xx/main.py
+++ b/src/SMU-Agilent_B29xx/main.py
@@ -68,11 +68,11 @@ class Device(EmptyDevice):
         self.pulse_mode = True  # enables pulsed signal option in GUI
         
         self.speed_types = {
-        "Very fast (0.01)": "0.01",
-        "Fast (0.1)": "0.1",
-        "Standard (1)": "1",
-        "Slow (10)": "10",
-        "Very Slow (100)": "100"
+            "Very fast (0.01)": "0.01",
+            "Fast (0.1)": "0.1",
+            "Medium (1)": "1",
+            "Slow (10)": "10",
+            "Very slow (100)": "100",
         }
 
     def set_GUIparameter(self):
@@ -88,7 +88,7 @@ class Device(EmptyDevice):
             "PulseOnTime": 1e-3,  # for use on the B2902B, defined as the pulse width time
             "PulseDelay": 0,  # for use on the B2902B, defined as the delay time prior to pulse
             "PulseOffLevel": 0.0,  # bias voltage during pulse-off
-            "PulseCount": 1, #amount of pulses
+            "PulseCount": 1,  # amount of pulses
             # "Average": 1, # not yet supported
         }
 
@@ -189,7 +189,7 @@ class Device(EmptyDevice):
             self.port.write(":SENS%s:FUNC \"VOLT\"" % self.channel)
             # measurement mode
             self.port.write(":SENS%s:VOLT:NPLC %s" % (self.channel, self.speed_types[self.speed]))
-            #NPLC definition for sensing voltage
+            # NPLC definition for sensing voltage
             self.port.write(":SENS%s:VOLT:PROT %s" % (self.channel, self.protection))
             # Protection with Imax
             self.port.write(":SENS%s:VOLT:RANG:AUTO ON" % self.channel)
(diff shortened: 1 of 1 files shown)


src/SMU-Agilent_B29xx/main.py
```diff
@@ -390,16 +399,43 @@ def apply(self):
...
     def call(self):
+        
+        if self.pulse:
+            opcounter = 0                                                    # set counter for operation register request loop back to zero
+                
+            while True:
+                self.port.write(":STAT:OPER:COND?")                           # query SMU for the status of the operation register
+                opstatus=self.port.read()
+                print("Operation Status:", opstatus)
+                if opstatus=="1170":                                          # checks whether transition and aquisition are finished on both channels (INT values 2+16 + 128+1024)
```
afischer-sweepme, line 430: Could it be that 1170 is achieved in your case, but other users where other operation status values might be added as well would have a different total number here. Would it make sense to check whether single bits have the correct state?

src/SMU-Agilent_B29xx/main.py
```diff
@@ -390,16 +399,43 @@ def apply(self):
...
+                    break
+                else:
+                    opcounter += 1
+                    time.sleep(0.5)
+  
+        
+        #if self.pulse:
```
afischer-sweepme, line 437: Commented out code: Still needed?

src/SMU-Agilent_B29xx/main.py
```diff
@@ -155,8 +162,8 @@ def configure(self):
         if self.source.startswith("Voltage"):
             self.port.write(":SOUR%s:FUNC:MODE VOLT" % self.channel)
             # sourcemode = Voltage
-            self.port.write(":SOUR%s:VOLT:MODE FIX" % self.channel)
-            # sourcemode fix
+            self.port.write(":SOUR%s:VOLT:MODE SWE" % self.channel)
```
afischer-sweepme, line 165: Would it make sense to use the Sweep mode only for the pulse while keeping it FIX for non-pulse measurements?

src/SMU-Agilent_B29xx/main.py
```diff
@@ -155,8 +162,8 @@ def configure(self):
...
+            # sourcemode sweepo
```
afischer-sweepme, line 166: typo sweepo

src/SMU-Agilent_B29xx/main.py
```diff
@@ -390,16 +399,43 @@ def apply(self):
...
+            if self.source.startswith("Voltage"):
+                self.port.write(":SOUR%s:VOLT:STAR %s" % (self.channel, value))
+                self.port.write(":SOUR%s:VOLT:STOP %s" % (self.channel, value))
+            
             # releasing the pulse trigger
             self.port.write(":INIT (@%s)" % self.channel)
+            
         else:
             # set output to specified values
             self.port.write(":SOUR%s:%s %s" % (self.channel, self.commands[self.source], value))
```
afischer-sweepme, line 419: I wonder that this works because the Sweep mode is now also set for non-pulse measurements. If Sweep mode would be only used for pulses, it would be more clear that this commands works as it is typically used with mode FIX.

src/SMU-Agilent_B29xx/main.py
```diff
@@ -390,16 +399,43 @@ def apply(self):
...
+            if self.source.startswith("Current"):
+                self.port.write(":SOUR%s:CURR:STAR %s" % (self.channel, value))
+                self.port.write(":SOUR%s:CURR:STOP %s" % (self.channel, value))
+            
+            if self.source.startswith("Voltage"):
+                self.port.write(":SOUR%s:VOLT:STAR %s" % (self.channel, value))
+                self.port.write(":SOUR%s:VOLT:STOP %s" % (self.channel, value))
+            
             # releasing the pulse trigger
             self.port.write(":INIT (@%s)" % self.channel)
```
afischer-sweepme, line 415: Releasing the pulse could be also shifted to a later phase. This can be beneficial when for example the measurement must be synchronized with other instruments. For example the Keithley 26xx driver is initiating the pulse in the "trigger_ready" phase just before "measure" phase.

Example: A motor and the pulsed SMU get a new set value during "apply". The motor needs some time to reach the new position. This is checked in the "reach" phase after all instruments had their "apply" phase. However, the SMU already sent the pulse during "apply". It would be not a problem in a DC measurement, but for pulsed measurements it is more important that the pulse is triggered after "reach" and "adapt" phase, just before the "measure".

For measurements with only one pulsed SMU there should be however no problem.

afischer-sweepme (commented):
Thanks for the further contribution. I have not much experience to say whether the Sweep is also a good choice for the single-spot DC measurement. One option would be to use FIX for the normal DC measurement and SWE for the pulsed measurments to allow for multiple pulses.

src/SMU-Agilent_B29xx/main.py
Code-Craftsman-Christian: Giving this a second though, it could well be than one is using channel A for a constant voltage and channel B in a consecutive SweepMe step to emit a pulse. In this case, transistion and aqusition on channel A would not be completed.

Proposal: check which channel the procedure is running on and check only for >=18 (CH1) and >=1152 (CH2)

afischer-sweepme: I did not read into how the operation status is organized, but yes it would make sense to do it for each selected channel individually and to check only those bits that contain relevant information.

src/SMU-Agilent_B29xx/main.py
afischer-sweepme: It would give me a better feeling if the standard DC measurement runs with FIX and the new pulse mode can run with SWE. I can help you with implementing it. Just let me know. Basically, I would shift this line to where the cases are for DC and pulse handling. In "apply" phase I would also add a section to handle DC and pulse mode.

afischer-sweepme (dismissed):
I saw that you pushed the latest changes in the "main" branch. As the first commit was in "patch-1" branch, I cherry-picked your latest commit and moved it to "patch-1" so that this pull request remains valid and we can proceed here with the merge.

I only did a bit cleanup of the comments. Most relevant change ist that I renamed speed type "Standard (1)" to "Medium (1") as this matches the wording of all other drivers and 
"Very Slow" -> "Very slow".

If you are fine with the changes, we can proceed with the merge.

Code-Craftsman-Christian (commented):
Thanks, looks fine.

afischer-sweepme (approved):
I have moved the pulse triggering to 'measure' as this phase is the one where a measurement should be initiated. If this does not work or makes problems later, we can resolve them in a new pull request. Sometimes some adaptions is needed until all drivers are perfectly harmonized.

PR #87, opened 2024-08-07
Keysight EXR scope driver

Code-Craftsman-Christian:
This driver was tested on a Keysight EXR scope, but MXR and UXR should work fine as well as only basic functionality was implemented and all three scopes run on the same software.

Development of the driver began with the Tek7k driver as a starting point but just few pieces are left. Beside configuration of channels, trigger and sample rate, the sample rate type option was misused to implement the possibility of ERES, offering the tradin of bandwidth and sample rate for an increase in ENOB, up to 16bit.

Some wait times had to be inserted to take care of processing time on the scope. Especially if you start the scope with default settings and request it to change pretty much everything from vertical scale to horizontal scale, trigger type, trigger edge, etc. etc. it could be that it triggers before finalizing on all changes, throwing an "invalid waveform" error because the captured one in memory does not fit the requested parameters.

The data transfer uses ASCii as it is nice to handle during debugging and requires not processing after retrival. If speed is a concern, a switch to WORD format including processing of the values needs to be implemented.

### DIFF
--- a/src/Keysight_EXR/license.txt
+++ b/src/Keysight_EXR/license.txt
@@ -1,0 +1,28 @@
+This Device Class is published under the terms of the MIT License.
+Required Third Party Libraries, which are included in the Device Class
+package for convenience purposes, may have a different license. You can
+find those in the corresponding folders or contact the maintainer.
+
+MIT License
+
+Copyright (c) 2021 SweepMe! GmbH
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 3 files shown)


src/Scope-Keysight_EXR/main.py
```diff
@@ -0,0 +1,326 @@
...
+             "TriggerSource": ["As is", "CHAN1", "CHAN2", "CHAN3", "CHAN4", "AUX", "LINE"],
+             # "TriggerCoupling": ["As is", "AC", "DC", "HF", "Auto level"],  # not yet implemented
+             "TriggerLevel": 0,
+             "TriggerDelay": 0,
+             "TriggerTimeout": 2,
+             "TimeRange": ["Time range in s", "Time scale in s/div"],
+             "TimeRangeValue": 5e-4,
+             "TimeOffsetValue": 0.0,
+             "SamplingRate": ["10e+3", "100e+3", "1e+6", "10e+6", "100e+6", "1e+9","5e+9","10e+9","16e+9"],
+             "SamplingRateType":["BITS10: 10 Bits (16 GSa or Manual)","BITS11: 11 Bits (6.4 GSa)","BITS12: 12 Bits (3.2 GSa)","BITS13: 13 Bits (1.6 GSa)","BITS14: 14 Bits (800 MSa)","BITS15: 15 Bits (400 GSa)","BITS16: 16 Bits (200 GSa)","BITS16_4: 16 Bits (100 GSa)","BITS16_2: 16 Bits (50 GSa)"],
```
afischer-sweepme, line 92: The field SamplingRateType can be used to present different options, e.g. as used in the Rohde&Schwarz RTE driver:
image

Best solution would be to introduce a new parameter like "ADC resolution" which we could do if you have no objections.

Code-Craftsman-Christian: This would be a welcome change, thanks Axel. I felt uncomfortable to use this field for the ADC resolution as the sampling rate value field lacks a proper descriptor now. But as I had a task at hand that required the higher ADC resolution, it fullfilled its need for the time being.

Code-Craftsman-Christian (commented):
Looks good, no objections.

afischer-sweepme:
@Code-Craftsman-Christian 
The Scope module now supports the key "ADCResolution". The driver is already adapted. I further changed the way how adc resolution options are defined (now via dictionary). Before we can merge, please test it once.
First you need to update the Scope module in the Version Manager. Then, you need to reload your setting to refresh the Scope module in your sequencer.

afischer-sweepme (approved):
There are no further requests. If the adc resolution works and we get your ok, we can merge

Code-Craftsman-Christian:
The ADC resolution feature works flawless, thanks. However, in attempt to fullfill requests presented to me by GitHub I've got failed checks alerts in this commit. Please see if this prevents a merge and if I can do something about it.

PR #93, opened 2024-08-30
HP 3456A DMM/Logger driver

Code-Craftsman-Christian:
Driver for the HP 3456A DMM, essentially a 6.5 Volt- and Ohmmeter, no current measurement capabilities. Besides DCV, ACV, DC+ACV, 2-Wire and 4-Wire resistance, the driver also includes the (O)ffset (C)ompensated measurement mode for 2W and 4W Resistance measurements. Number of NPLC, Auto-Zero and analogue Filter options are available, too. The display option does not actually switch of the display but just switches every digit to display an uppercase "H" to hide the values. No idea if this is of use, but it is implemented as a native instrument function anyway.

Note a few specialities: 
- this is pre-SCPI, the HP-IB commands are few and short. 
- there is no command to request a measurement, rather the current result can be retrieved from the instrument just by reading the bus
- the device has a hardware-enabled talking-only mode to feed a printer or similar; watch out that it is not set as it won't accept commands via GPIB while in this mode, which is set via the GBPI address dip switches (see driver comment)

The 3456A is known for its superb long-time stability and used even today as a lab-internal reference in many places.

### DIFF
--- a/src/Logger-HP_3456A/main.py
+++ b/src/Logger-HP_3456A/main.py
@@ -35,9 +35,7 @@ from pysweepme.EmptyDeviceClass import EmptyDevice
 
 
 class Device(EmptyDevice):
-    description = """<p><strong>HP 3456A 6.5 digit ohm- and voltmeter</strong>
-                     <br>Make sure the instrument is NOT in listen-only mode:
-                     <br>rightmost DIP switch of GPIB address encoder must be in "down" position </p>
+    description = """<p><strong>HP 3456A 6.5 digit ohm- and voltmeter</strong></p></p>
                     """
 
     def __init__(self):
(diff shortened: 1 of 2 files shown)


src/Logger-HP_3456A/main.py
```diff
@@ -0,0 +1,196 @@
...
+        pass
+
+    def configure(self):
+
+        # Mode
+        self.port.write("%s" % self.modes[self.mode])
+
+        # Resolution
+        print(self.resolution)
+        print(self.mode)
```
franz-sweepMe, line 155: If these debugs/prints are necessary, consider adding a `debug: bool = True` parameter to the GUI

franz-sweepMe (commented):
A very clean, well-written driver. Thanks for your contribution. 
I already added the license.txt file. Once you have resolved the comment on the print command, I would like to run our auto-formatting before merging.

src/Logger-HP_3456A/main.py
afischer-sweepme: In future there might be a general debug option available that tells a driver to be verbose. From this point of view I would rather not add a debug option to the GUI, but just remove the prints to have a clean silent driver and whenever new featues are added, one can add new print statement to check something.

src/Logger-HP_3456A/main.py
```diff
@@ -0,0 +1,196 @@
...
+        # Display
+        if self.display == "Off":
+            self.port.write("D0")
+
+    def unconfigure(self):
+        if self.display == "Off":
+            self.port.write("D1")  # We switch Display on again if it was switched off
+
+    def measure(self):
+        self.data = self.port.read()  # retrieves current measurement data from the instrument
```
afischer-sweepme, line 193: I wonder how the measurement is triggered. Is the instrument permanently sending data? Could it be that there is old data in the buffer or is the readout somehow triggering a fresh measurement. Maybe adding a further comment?

src/Logger-HP_3456A/main.py
Code-Craftsman-Christian: Yes, please remove. I like to have the IDs print out in my drivers, but they should not be included in the public drivers and I always forget to remove them.

src/Logger-HP_3456A/main.py
Code-Craftsman-Christian: The standard setting uses the internal trigger which delivers the result to the output buffer. Once it is read, any following request has to wait for a new result. This is easily visible during normal use: raising the NPLC will increase the time between results proportionally as SweepMe has to wait for an answer until the integration is finished.

There is no specific measurement retrieval command, we just could use a software trigger instead of the internal trigger to conduct a measurement on the device. Retrieving it would be the same, however. But if you would rather promote having a manual trigger of the measurment for a timed start within e.g. the 'measure' procedure, I could implement it and try out.

afischer-sweepme: If there is a software trigger, I would recommend to use the software trigger during 'measure' as this starts the measurement and the readout of the result can be done during 'read_result' phase or also possible within 'measure' phase after the trigger. This ensures to have always a most recent meaurement in the buffer.
Using 'read_result' phase for readout has the advantage that you could use multiple multi-meters at the same time. They get all software triggered during 'measure' phase and the results can be read out one phase later.

A problem with the internal trigger can occur if you e.g. use multiple branches or a long delay and the buffer fills with data and the next time measure is used, an old data set is read out.

Sometimes, one needs to set the trigger back to internal after the measurement to see an update of the values on the display of the instrument again.

Code-Craftsman-Christian: Alright, I will adjust the driver next week and update my branch accordingly. No need to pull it prior to the changes.

Thanks for the hint regarding the switchback to the internal trigger, makes sense.

Code-Craftsman-Christian: New revision commited.

franz-sweepMe (approved):

afischer-sweepme (approved):

PR #96, opened 2024-09-03
Initial driver release HP 3457A DMM

Code-Craftsman-Christian:
Initial driver release for the HP 3457A DMM, successor to the previously released 3456A. The 3457A added the support for current measurement. In addition, it has a trick uppon his sleeve: despite being a 6.5 digit DMM, a 7th digit can be retrieved via GPIB. This comes in handy when making averaged readings on a constant value where the native noise gets reduced and the additional digit can be of real use. Therefore, the option was implemented.

The unit was introduced before SCPI so it uses an HP-IB based command set which is kind of in-between what the 3456A uses and what can be seen in newer units based on SCPI.

### DIFF
--- a/src/Logger-HP_3457A/main.py
+++ b/src/Logger-HP_3457A/main.py
@@ -1,0 +1,206 @@
+# This Device Class is published under the terms of the MIT License.
+# Required Third Party Libraries, which are included in the Device Class
+# package for convenience purposes, may have a different license. You can
+# find those in the corresponding folders or contact the maintainer.
+#
+# MIT License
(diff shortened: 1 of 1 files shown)


franz-sweepMe (approved):
A well-commented driver, thanks for your contribution!

For future PRs, you can add a license.txt file to the driver folder. For drivers consisting only of a main.py file, this is not strictly necessary, as the license is already added in the main.py header. But it helps to keep the repository consistent. 
I added the license.txt and ran the auto-formatting script.

PR #100, opened 2024-10-28
SMU-Keithley 4200-SCS: Add Pulse Mode and List Mode

franz-sweepMe:
Add Pulse mode and List Mode to the SMU 4200-SCS driver.
Add a more precise check to decide if the driver is run on the hardware or via lptlib server.
Update formatting, docstrings, and type annotations to the newest standard.


src/SMU-Keithley_4200-SCS/main.py
```diff
@@ -109,26 +138,84 @@ def __init__(self):
...
+            "time": "time",
+        }
+        """The keys are used to register and receive the list sweep values."""
+
+        self.list_delay_time: float = 0.0
+
+    @staticmethod
+    def find_ports() -> list[str]:
+        """Find available ports."""
+        return ["LPTlib"] if RUNNING_ON_4200SCS else ["LPTlib via xxx.xxx.xxx.xxx"]
```
fk3, line 206: Is that what's shown to the user? Or is xxx.... somehow replaced with the actual IP Address?
More user friendly alternative would be "LPTlib via IP" or "LPTlib via TCPIP" or "LPTlib via network"

franz-sweepMe: The user has to exchange the xxx with the IP address of the server.exe

fk3 (approved):

PR #101, opened 2024-10-29
E3631(!)A: handover of self.value as float + software averaging function 

Code-Craftsman-Christian:
Bugfix: self.value is now handed over explizitly as float. This fix was necessary for the "Test Field" function to work properly as contrary tp values out of the sweep list, the text field hands over its content in string format.

Added Feature: the SMU class offers the option of averaging values. The E36xxA instruments offer a read-back resolution (model dependend) down to 1mA, often saving the use of an additional DMM. By sampling multiple measurement values, the readout can be stabilized conderable, making these PSUs suitable for recording DC curves on their own for many simple parts. This is now implemented, offering the option of averaging with a simple click of an options box within the SMU class GUI.

### DIFF
--- a/src/SMU-Keysight_E3631A/main.py
+++ b/src/SMU-Keysight_E3631A/main.py
@@ -4,7 +4,7 @@
 # find those in the corresponding folders or contact the maintainer.
 #
 # MIT License
-# 
+#
 # Copyright (c) 2024 SweepMe! GmbH (sweep-me.net)
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -36,8 +36,7 @@ from pysweepme.EmptyDeviceClass import EmptyDevice
 
 
 class Device(EmptyDevice):
-    """
-    description =
+    """description =
     <p><strong>Notes:</strong></p>
     <ul>
     <li>COM Port untested as of 20240807</li>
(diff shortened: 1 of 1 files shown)


src/SMU-Keysight_E3631A/main.py
```diff
@@ -159,35 +165,42 @@ def poweroff(self):
         self.port.write("OUTP:STAT OFF")
 
     def apply(self):
```
franz-sweepMe, line 167: instead of checking `float(self.value)` in every `if` clause, you could write `self.value = float(self.value)` in the first line of `apply`

src/SMU-Keysight_E3631A/main.py
```diff
@@ -88,6 +89,11 @@ def get_GUIparameter(self, parameter={}):
         self.source = parameter["SweepMode"]
         #self.route_out = parameter["RouteOut"]
         self.currentlimit = parameter["Compliance"]
+        self.average = int(parameter["Average"])
+        
+        if self.average == 0:
```
franz-sweepMe, line 94: Same comment as in #102, you can check for average < 1

franz-sweepMe (commented):
Nice work, thanks for contributing!

franz-sweepMe (approved):

PR #102, opened 2024-10-29
E3632(!)A: handover of self.value as float + software averaging function

Code-Craftsman-Christian:
Bugfix: self.value is now handed over explizitly as float. This fix was necessary for the "Test Field" function to work properly as contrary tp values out of the sweep list, the text field hands over its content in string format.

Added Feature: the SMU class offers the option of averaging values. The E36xxA instruments offer a read-back resolution (model dependend) down to 1mA, often saving the use of an additional DMM. By sampling multiple measurement values, the readout can be stabilized conderable, making these PSUs suitable for recording DC curves on their own for many simple parts. This is now implemented, offering the option of averaging with a simple click of an options box within the SMU class GUI.

### DIFF
--- a/src/SMU-Keysight_E3632A/main.py
+++ b/src/SMU-Keysight_E3632A/main.py
@@ -75,6 +75,7 @@ class Device(EmptyDevice):
             "RouteOut": ["Front"],
             "Compliance": 1,
             "RangeVoltage": ["15 V / 7 A", "30 V / 4 A"],
+            "Average": 1,
         }
 
         return gui_parameter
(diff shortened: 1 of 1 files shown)


src/SMU-Keysight_E3632A/main.py
```diff
@@ -90,7 +91,12 @@ def get_GUIparameter(self, parameter={}):
         # self.channel = int(parameter["Channel"])
 
         self.voltage_range = parameter["RangeVoltage"]
-
+        self.average = int(parameter["Average"])
+        
+        if self.average == 0:
+            msg = ("Average of 0 not possible. Disable average by setting it to 1.")
```
franz-sweepMe, line 97: The SMU GUI input for "Average" only allows values >1. It still makes sense to check for incorrect values of self.average (e.g. when the driver is used with pysweepme in a Python script instead of the SweepMe! GUI), but it could be generalized to check for self.average < 1

franz-sweepMe (commented):
The changes are useful contributions to the driver, thanks for you work @Code-Craftsman-Christian

src/SMU-Keysight_E3632A/main.py
Code-Craftsman-Christian: Agreed, a change to self.average < 1 is sensible.

franz-sweepMe (approved):

PR #103, opened 2024-11-11
enhance NGU compatibility

Code-Craftsman-Christian:
The original driver caused two SCPI errors that have been fixed by identifying the model via "*ID?" and adjusting the code accordingly via IF statements.

Modifications for NGU models:
- fastest NPLC is defined as "0" instead of "0.1" and needed for its "FastLog" capability
- the ""INST:OUT" command for selecting an output is not only unnecessary as this model is a single channel SMU; it is also rejected by the NGU in form of an SCPI command error. While the command is mentioned in the handbook in just one location within an example code, the command itself is not included.

The driver worked flawless without these modifications and all errors might have been went by unnoticed if the SCPI error beep was deactivated and the fastest NPLC setting was never used.

### DIFF
--- a/src/SMU-Rohde&Schwarz_NGx/license.txt
+++ b/src/SMU-Rohde&Schwarz_NGx/license.txt
@@ -5,7 +5,7 @@ find those in the corresponding folders or contact the maintainer.
 
 MIT License
 
-Copyright (c) 2022-2023 SweepMe! GmbH (sweep-me.net)
+Copyright (c) 2022-2024 SweepMe! GmbH (sweep-me.net)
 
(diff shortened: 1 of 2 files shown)


afischer-sweepme (commented):
I pushed some minor changes and updated the license. It is a pity that the NGU model does not understand the set channel command and how it handles the nplc. I would like to take a look into the manual and maybe contact R&S, but you could already have a look whether you are happy with my changes.

Code-Craftsman-Christian:
It's all well done and dusted, thanks Axel.

fk3 (approved):

PR #107, opened 2024-11-25
aSpect Systems idSMU Driver

franz-sweepMe:
Add an SMU driver for the aSpect Systems idSMU2.
The driver supports
- auto ranging
- List Sweep
- List Sweep for multiple channels

Currently, the driver supports current ranges and channel selection only for model idSMU2.
The driver has been tested on a single SMU. Board configurations with multiple SMU modules have not been tested yet.


fk3 (commented):
Looking at the libs folder, the driver only supports Python 3.9 64bit? In that case adding architecture information in the config.ini would make sense

fk3 (approved):

PR #109, opened 2024-11-28
Switch-CTS_CS: working version

afischer-sweepme:
Driver for CTS climate chambers: Tested and working version

### DIFF
--- a/src/Switch-CTS_CS/main.py
+++ b/src/Switch-CTS_CS/main.py
@@ -5,7 +5,7 @@
 #
 # MIT License
 # 
-# Copyright (c) 2021 SweepMe! GmbH
+# Copyright (c) 2024 SweepMe! GmbH (sweep-me.net)
 # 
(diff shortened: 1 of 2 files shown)


src/Switch-CTS_CS/main.py
```diff
@@ -335,16 +312,23 @@ def configure(self):
                             
         else:
             self._temperature_set_value = None
+
+        self.reach_humidity = self.reach_humidity == "1" or self.reach_humidity == "True" or self.reach_humidity is True
```
fk3, line 316: I think it would make sense put the "how to interpret input values" logic into the input handling function, i.e. get_GUIparameter(), like
```
self.reach_humidity = parameter["Reach humidity"] in ["1", "True", True]
```
Then self.reach_humidity will always be a bool everywhere in the driver and does not require to think about if it was already converted or not

fk3 (approved):

PR #110, opened 2024-12-05
Switch keithley 707 x

franz-sweepMe:
Add a Driver for the Switching Matrix Keithley 707A. The driver has been tested with the device.
Because the 707A is an older model the commands are incompatible with the existing 707B driver.
Clean up of the 707B driver (formatting + docstrings), but no changes to the functionality as it could not be tested.


src/Switch-Keithley_707B/main.py
```diff
@@ -34,158 +34,147 @@
...
+        self.savetype = [True]
 
-    def set_GUIparameter(self):
-        
-        GUIparameter = {
+        # Measurement parameters
+        self.sweepmode: str = "Channels"
+        self.switch_settling_time_s: float = 0.004
+
+    @staticmethod
```
fk3, line 84: set_GUIparameter should not be a staticmethod. Even though it does not need `self`, the parent class (EmptyDevice) defines set_GUIparameter to be a non-static method and child classes should use the same signature

src/Switch-Keithley_707A/license.txt
```diff
@@ -0,0 +1,30 @@
+This Device Class is published under the terms of the MIT License.
+Required Third Party Libraries, which are included in the Device Class
+package for convenience purposes, may have a different license. You can
+find those in the corresponding folders or contact the maintainer.
+
+MIT License
+
+Copyright (c) 2021, 2024 SweepMe! GmbH (sweep-me.net)
```
fk3, line 8:
```suggestion
Copyright (c) 2024 SweepMe! GmbH (sweep-me.net)
```

src/Switch-Keithley_707A/main.py
```diff
@@ -0,0 +1,145 @@
...
+
+        self.variables = []
+        self.units = []
+        self.plottype = []
+        self.savetype = []
+
+        #     Measurement parameters
+        self.sweepmode: str = "Channels"
+
+    @staticmethod
```
fk3, line 74: don't make set_GUIparameter static, see comment on 707B driver

src/Switch-Keithley_707A/main.py
```diff
@@ -0,0 +1,145 @@
...
+        self.port_types = ["GPIB"]
+        self.port_properties = {
+            "timeout": 5,
+            "EOL": "\r",
+        }
+
+        self.variables = []
+        self.units = []
+        self.plottype = []
+        self.savetype = []
```
fk3, line 69: What's the reason / difference between 707A/B why this one does not have any variables and the other drives has a channel variable? (Though channel does not really seem to be a channel but rather the Switching Matrix state?)

franz-sweepMe: The 707B returns self.value, which is a comma-separated string of all channels that are set to closed in apply. 
I was unsure if this information is of any use, as it is the same as the input value. To remain backward compatible, I did not change the return values of the 707B.
For 707A, we could return the input string as well (same problem, I do not think this is particular helpful) or we could query the channel state of the matrix and return the complete status as comma-separated string or any other data format. As these devices have typically >8x36 channels, the return value would be very large and the usefulness of this info is unclear.

franz-sweepMe: If we (later) add a sweep mode where the previous state of the matrix is not reset in apply, but rather updated, it could be helpful to return the updated matrix state for double-checking

franz-sweepMe: Return of self.value has been added to link the matrix state to the measurement result data.

fk3: After adding self.value to the return value of call(), self.variables, ...units, etc. also need to be updated. Otherwise SweepMe! expects zero variables, but gets one value from call()

fk3 (approved):

PR #111, opened 2024-12-05
LCR meter Keithley 590

franz-sweepMe:
Add an LCR Meter driver for Keithley 590.
- The driver has been tested with the device.
- The averaging functionality is currently not supported, as it is unclear if device-side averaging works for non-sweep measurements.
- The device returns capacitance, conductance, and voltage. The driver could convert these values into the resistance and reactance, so that the SweepMe LCRmeter module can calculate capacitance, conductance, ... again. It is open for discussion if this is useful

### DIFF
--- a/src/LCRmeter-Keithley_590/main.py
+++ b/src/LCRmeter-Keithley_590/main.py
@@ -43,10 +43,10 @@ class Device(EmptyDevice):
         """Initializes the device class."""
         super().__init__()
 
-        self.variables = ["R", "X", "Frequency", "Voltage bias"]
-        self.units = ["Ohm", "Ohm", "Hz", "V"]
-        self.plottype = [True, True, True, True]
-        self.savetype = [True, True, True, True]
+        self.variables = ["C", "G", "Voltage bias"]
+        self.units = ["F", "S", "V"]  # TODO: Check units
+        self.plottype = [True, True, True]
+        self.savetype = [True, True, True]
 
         self.port_manager = True
         self.port_types = ["GPIB"]
@@ -70,10 +70,6 @@ class Device(EmptyDevice):
         }
 
         self.frequency: str = "100kHz"
-        self.frequencies = {
-            100E6: "F0",
-            1E9: "F1",
-        }
 
         self.trigger: str = "One-shot, talk"
         self.trigger_modes = {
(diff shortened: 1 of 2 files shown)


src/LCRmeter-Keithley_590/main.py
```diff
@@ -0,0 +1,335 @@
...
+        self.frequency = "1MHz" if frequency == 1E9 else "100kHz"
+
+        self.reading_rate = parameter["Integration"]
+        self.average = 1  # int(parameter["Average"])  # uncomment if averaging is needed for non-sweep mode
+
+        self.measure_range = parameter["Range"]
+        self.trigger = parameter["Trigger"]
+
+    def initialize(self) -> None:
+        """Initialize the Keithley 4200-SCS LCRmeter."""
```
fk3, line 140: copy/paste leftover?

src/LCRmeter-Keithley_590/main.py
```diff
@@ -0,0 +1,335 @@
...
+            msg = f"Voltage {voltage} is out of range. Choose between -20 V and 20 V."
+            raise ValueError(msg)
+
+        # single voltage
+        command = f"V,,,{voltage},{averaging}X"  # Commas needed to skip start, stop, and step values
+        self.port.write(command)
+
+    def set_waveform(self) -> None:
+        """0 DC, 1 Single Staircase, 2 Dual Staircase, 3 Pulse Train, 4 External bias source."""
+        waveform = 1
```
fk3, line 289: Should this function still get an argument? Otherwise the docstring does not really describe how the function is used, but it is rather an internal comment

src/LCRmeter-Keithley_590/main.py
```diff
@@ -0,0 +1,335 @@
...
+        start = 1  # s
+        stop = 10
+        step = 0.1
+        self.port.write(f"W {waveform},{start},{stop},{step}X")
+
+    def set_bias_on(self, bias_on: bool = True) -> None:
+        """Turn the voltage on."""
```
fk3, line 296:
```suggestion
        """Turn the voltage on/off."""
```

src/LCRmeter-Keithley_590/main.py
```diff
@@ -0,0 +1,335 @@
...
+
+    def set_trigger_mode_and_source(self, trigger: str) -> None:
+        """Set the trigger mode and source."""
+        command = f"T{self.trigger_modes[trigger]}X"
+        self.port.write(command)
+
+    def set_zero_on(self, set_on: bool = True) -> None:
+        """Set the zero."""
+        command = "Z1X" if set_on else "Z0X"
+        self.port.write(command)
```
fk3, line 264: What does it mean to "set zero"? And it is not required to set it off again after the measurement / unconfigure?

fk3 (approved):

PR #113, opened 2024-12-12
Logger Rohde&Schwarz RTX

franz-sweepMe:
Add a logger driver for Rohde&Schwarz RT-Series Oscilloscopes.
These devices allow users to set "measurement places" that extract specific values such as frequency, peak height, rise time, ... from the measured signal.

The driver has been tested on RTB2004 and should also work on other RT Oscilloscopes as they share the same commands.

### DIFF
--- a/src/Logger-Rohde&Schwarz_RTX/main.py
+++ b/src/Logger-Rohde&Schwarz_RTX/main.py
@@ -54,7 +54,7 @@ class Device(EmptyDevice):
         self.port_manager = True
         self.port_types = ["GPIB", "TCPIP", "USB"]
         self.port_properties = {
-            "timeout": 2.0,
+            "timeout": 20,  # higher timeout if noise is measured, some mode need more time
         }
 
         # SweepMe return parameters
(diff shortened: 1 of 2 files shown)


src/Logger-Rohde&Schwarz_RTX/main.py
```diff
@@ -0,0 +1,252 @@
...
+                    places are measurements such as as Maximum peak height, periodicity, rise time, and much more. The
+                    measurement places can be definied on the device or via SweepMe! for given mode and source
+                    (e.g. Channel 1-4).
+                    <p>The driver has been tested with RTB2004, but should also work with other RT-Series devices.</p>
+                    </p>
+                    <p>Setup:</p>
+                    <ul>
+                    <li>Enable remote control at device.</li>
+                    <li>If use preset, the defined measurement places from the devices are used. Note that currently
+                    SweepMe! is unable to display the used measurement mode and units in preset mode.</li>
```
fk3, line 51: Is that really a limitation of SweepMe! or rather this driver?

franz-sweepMe: The label and unit of the measurement, which correspond to self.variables and self.units, can only be requested after a connection to the device has been made. AFAIK SweepMe needs to know the variables and units before connect. I was unable to expand/change the variables afterwards

afischer-sweepme (approved):

PR #115, opened 2024-12-19
Signal Rohde&Schwarz SMB100A

franz-sweepMe:
Add a signal generator driver for Rohde&Schwarz SMB100A
The driver has been tested with the device.

Before merging, the driver will be tested on SMC100A to see if the driver can be generalized.

### DIFF
--- a/src/Signal-Rohde&Schwarz_SMB100A/main.py
+++ b/src/Signal-Rohde&Schwarz_SMB100A/main.py
@@ -54,14 +54,15 @@ class Device(EmptyDevice):
 
         # Measurement Parameters
         self.sweep_mode: str = "None"
-        self.frequency_input: str = "Frequency in Hz"  # or "Period in s"
-        self.frequency: float = 2.2e9
+        self.period_frequency_mode: str = "Frequency in Hz"  # or "Period in s"
+        self.period_frequency_value: float = 2.2e9
 
-        self.amplitude_mode: str = "Amplitude in V"  # or "High level in V"
+        self.amplitude_modes = ["Amplitude in V", "Amplitude in dBm", "Amplitude in dBuV"]
+        self.amplitude_unit: str = "V"  # or "High level in V"
         self.amplitude: float = 0.1
 
-        self.delay_mode: str = "Phase in deg"  # or "Delay in s"
-        self.delay: float = 0.0
+        self.delay_phase_mode: str = "Phase in deg"  # or "Delay in s"
+        self.delay_phase_value: float = 0.0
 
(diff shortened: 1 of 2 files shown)


src/Signal-Rohde&Schwarz_SMB100A/main.py
```diff
@@ -0,0 +1,329 @@
...
+        }
+
+    def get_GUIparameter(self, parameter: dict) -> None:  # noqa: N802
+        """Get the GUI parameters."""
+        self.sweep_mode = parameter["SweepMode"]
+        self.frequency_input = parameter["PeriodFrequency"]
+        self.frequency = float(parameter["PeriodFrequencyValue"])
+
+        self.amplitude_mode = parameter["AmplitudeHiLevel"]
+        self.amplitude = float(parameter["AmplitudeHiLevelValue"])
```
afischer-sweepme, line 95: Because it is not clear yet whether it is an amplitude, the variable could be also called self.amplitude_highlevel_value. 
The mode could be called "self.amplitude_highlevel_mode".

src/Signal-Rohde&Schwarz_SMB100A/main.py
```diff
@@ -0,0 +1,329 @@
...
+
```
afischer-sweepme, line 96: No support for Offset and Low level?

src/Signal-Rohde&Schwarz_SMB100A/main.py
```diff
@@ -0,0 +1,329 @@
...
+        # self.port.write("*CLS")
+
+    def configure(self) -> None:
+        """Configure the device."""
+        self.lock_user_interface(True)
+
+        self.set_power_mode("CW")
+        self.set_power_unit("V")
+
+        self.set_power_level(self.amplitude)
```
afischer-sweepme, line 115: This line suggest that we handover the amplitude, but it is still unclear whether the value is interpreted as amplitude or high level which is why it would make sense to give a variable name that reflects this. The function name set_power_level could be also changed to "handle_amplitude_highlevel" to make clear that is not known yet whether the power is set, especially because it also not a power that is set but rather a voltage.

afischer-sweepme (commented):
Here is a good example how to handle amplitude, high level, offset, low level, frequency, period, delay and phase:
https://github.com/SweepMe/instrument-drivers/blob/main/src/Signal-Siglent_SDG2000X/main.py

Is it possible to design the driver in a similar way?

afischer-sweepme (approved):

PR #133, opened 2025-02-06
Updated driver HP 3456A

Code-Craftsman-Christian:
Noticed some minor flaws and fixed them.

- corrected an if-statement that was still not working due to the previous mix-up of "range" and "resolution"
- removed old code segments in comments
- added external trigger option
- additional comments
- formatting comments
- added additional NPLC speed settings after consultation of manual


franz-sweepMe (approved):
Thanks for your contribution!

PR #134, opened 2025-02-06
Update Keysight 34401A Driver

Code-Craftsman-Christian:
- NPLCs were added as dictionary but not used as such until now
- added additional NPLC speed
- corrected IF statement that was still flawed by the mix-up of "range" and "resolution" that was supposed to be fixed last time
- formatting of comments
- additional comments


franz-sweepMe (approved):
Thanks for your contribution!

PR #141, opened 2025-02-11
SMU Keithley 4200-SCS

franz-sweepMe:
- add custom list sweep and custom delay list
- update for lptlib server 1.1.2
- add error handling


src/SMU-Keithley_4200-SCS/main.py
```diff
@@ -645,17 +659,53 @@ def trigger_ready(self) -> None:
...
     def measure(self) -> None:
         """Start the pulse or list measurements. This cannot be done in 'apply' as the sweep value does not change."""
         if self.pulse_mode and self.pulse_master:
             # TODO: only the master driver instance needs to execute
             self.lpt.pulse_exec(mode=self.param.PULSE_MODE_SIMPLE)  # Alternatively self.param.PULSE_MODE_ADVANCED
 
         if self.list_master:
+            # Check if all elements of list_sweep_values are float
+            if not all(isinstance(value, float) for value in self.list_sweep_values):
+                msg = "Unsupported type detected in list values. Make sure all values are float"
```
fk3, line 694: Is it even possible that the elements are something different than float?
handle_list_sweep_parameter() seems to make sure of that already.

If this check is some kind of second line of defense, the error message should make clear that it's a bug in the driver. Because the SweepMe! User can't do anything to "make sure all values are float", right?

fk3: Probably same for list_delay_list below?

fk3 (approved):

PR #147, opened 2025-02-28
Bugfix 20250228

Code-Craftsman-Christian:
- important bugfix, see comments regarding undocumented CLE command behaviour; memory waveform gets cleared now.
- timebase offset command was forgotten to be implemented, sorry

### DIFF
--- a/src/Scope-Rigol_DHOseries/main.py
+++ b/src/Scope-Rigol_DHOseries/main.py
@@ -275,7 +275,7 @@ class Device(EmptyDevice):
             # will execute either :RUN oder :SINGle SCPI command depending on choice of aquisition type (continuous or single)
             self.port.write("%s" % self.aquisitiontypes[self.aquisitiontype])
 
-        time.sleep(1)
+        time.sleep(0.4)
         # The RIGOL does not allow to determine the amount of successful acquired averaging samples.
         # It also returns TD for trigger status during continuous triggering when a trigger was successful, but already waiting for a new one.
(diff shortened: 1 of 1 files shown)


Code-Craftsman-Christian:
addendum: third bugfix by adding a time.sleep after trigger mode setting as it can happen that the device still reports "STOP" trigger status afterwards if the status request is send out too fast after the tigger mode setting command.

@franz-sweepMe can you please implement the pull request with a modified time.sleep() in line #276? 1s was a fail-safe time I used at the beginning. I tested it down to 0.4 meanwhile (0.2s does not work, 0.3s does, so 0.4s should be save). I cannot find the option to cancel the pull request and restart an updated one.

src/Scope-Rigol_DHOseries/main.py
```diff
@@ -30,16 +30,15 @@
 # * Instrument: Rigol DHO1000 Series
 
 
-from pysweepme.EmptyDeviceClass import EmptyDevice
+from EmptyDeviceClass import EmptyDevice
```
franz-sweepMe: using `from pysweepme.EmptyDeviceClass import EmptyDevice` is preferred to denote that the drivers cannot be used without pysweepme (Either a SweepMe! main software installation or the pysweepme Python package are needed)

franz-sweepMe:
Thanks for your contribution @Code-Craftsman-Christian!
I think you can make changes to an existing PR by committing your changes to the branch you propose to merge.

franz-sweepMe (approved):

PR #159, opened 2025-04-28
Keithley 2700

franz-sweepMe:
Update the measurement step to use trigger_ready instead

### DIFF
--- a/src/Logger-Keithley_2700/main.py
+++ b/src/Logger-Keithley_2700/main.py
@@ -130,8 +130,6 @@ class Device(EmptyDevice):
         self.units=["Unit1"]
         self.plottype=[True]
         self.savetype=[True]
-        self.had_trigger = False
-
 
     def set_GUIparameter(self) -> dict:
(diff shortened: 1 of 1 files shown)


src/Logger-Keithley_2700/main.py
```diff
@@ -265,30 +267,57 @@ def configure(self):
     def deinitialize(self):
         self.port.write("SYST:BEEP:STAT ON")  # control-Beep on
 
-    def measure(self):
-        self.port.write("form:elem READ\n;READ?")
+    def trigger_ready(self):
+        self.had_trigger = False
+        self.port.write("INIT") #initialize trigger
```
afischer-sweepme, line 272: Triggering the measurement should be moved to 'measure' as this is the step where all instrument should start with the data acquisition. So it should be fine if 'trigger_ready' is replaced with 'measure'

afischer-sweepme (changes requested):

PR #160, opened 2025-05-06
Create SpectrumAnalyzer-HP_859xE

Code-Craftsman-Christian:
Initial driver release for HP 859xE Spectrum Analyzers. Might work for 856xA/E and 859xA as well but untested.


src/SpectrumAnalyzer-HP_859xE
```diff
@@ -0,0 +1,241 @@
...
+        # erase TRACE A to C; will also remove max/min hold function if previously enabled
+        self.port.write("BLANK TRC")
+        self.port.write("BLANK TRB")
+        self.port.write("BLANK TRA")
+        
+        # if setup_cmd_a starts with "CF" it means frequencies were entered as Center and Span and a frequency array from center-span/2 to center+span/2 with 401 steps is created
+        if self.setup_cmd_a.startswith("CF"):
+            self.frequencies = np.linspace(float(self.fv1) - float(self.fv2)/2, float(self.fv1) + float(self.fv2)/2, 401)
+        # if setup_cmd_a does not start with "CF" it means frequencies were entered as start and stop frequencies and a frequency array from start to stop with 401 steps is created
+        else:
```
franz-sweepMe, line 184: initialize() should contain only functions to prepare the instrument once. Dynamic parameters such as the frequency range, which could change during the sweep (e.g. by implementing dynamic ranges via the ParameterSyntax) should be handled in configure. Otherwise, the changes will not be applied.

src/SpectrumAnalyzer-HP_859xE
```diff
@@ -0,0 +1,241 @@
...
+        self.port.write("CLRW TRA") # makes sure Trace A is the active one
+        self.port.write("CONTS") # return to continuous sweep upon exit
+        
+    def measure(self):
+        # set to single sweep
+        self.port.write("SNGLS")
+        
+        # in case the standard aquisition mode "Current" is used, we ignore the repetitions set via GUI
+        if self.trace_mode == "Current":
+            self.port.write("CLRW TRA") # makes sure Trace A is the active one
```
franz-sweepMe, line 220: in initialize(), all traces were reset. Here, only Trace A is used. Would it be possible to add the trace as GUI parameter to choose which trace to use?
To prevent erasing a trace used by another instance of the driver, only the chosen trace should be cleared in initialize.

src/SpectrumAnalyzer-HP_859xE
```diff
@@ -0,0 +1,241 @@
...
+            for n in range (0, self.repetitions):
+                self.port.write("TS") # record sweep
+
+        if self.trace_mode == "Min hold":
+            self.port.write("CLRW TRA") # makes sure Trace A is the active one
+            self.port.write("MINH TRA") # maximum hold for Trace A
+            for n in range (0, self.repetitions):
+                self.port.write("TS") # record sweep
+
+        self.port.write("TRA?") # transfer Trace A in 16bit
```
franz-sweepMe, line 236: The measure() step should be used to start the measurement. Retrieving the results should be moved to read_result. See https://wiki.sweep-me.net/wiki/Sequencer_procedure for reference

src/SpectrumAnalyzer-HP_859xE
```diff
@@ -0,0 +1,241 @@
...
+        
+        # setting resolution bandwidth
+        self.port.write("RB %s" % self.rbws[self.rbw])
+        
+        # setting video bandwidth
+        self.port.write("VB %s" % self.vbws[self.vbw])
+        
+        # enable/disable video average function
+        if self.vavg == True:
+            self.port.write("VAVG ON")
```
franz-sweepMe, line 202: If self.vavg is False, should the VAVG OFF command be sent?

franz-sweepMe (changes requested):
Thanks for your contribution!

Please create a folder named "SpectrumAnalyzer-HP_859xe" in src, move this file there, and rename it to "main.py" to adhere to the SweepMe! driver template.

src/SpectrumAnalyzer-HP_859xE
Code-Craftsman-Christian: Certain functions are restricted to some specific Traces. I started to implement those functions, which is why I needed to make sure that all three traces are cleared. But in the end, this did not work out the way I expected them to do, which is why I removed them for the time being but forgot to update the trace clearings. I edited the code to just take care of Trace A, leaving the rest untouched.

src/SpectrumAnalyzer-HP_859xE
Code-Craftsman-Christian: Looks like I accidentially deleted the line with the function call. Fixed now.

src/SpectrumAnalyzer-HP_859xE
Code-Craftsman-Christian: Yes, makes sense. Added to the code.

Code-Craftsman-Christian:
I've created the new main.py file with all changes and will start a pull request once it had a test run on the instrument.

PR #227, opened 2026-01-27
Robot Thorlabs NanoTrak

franz-sweepMe:
Add driver for fiber auto-alignment controller NanoTrak.
The main routine of tracking-latching is done in configure(). The device does not really measure anything, but uses its power reading for optimization.


fk3 (commented):
/gemini review

fk3 (approved):

PR #230, opened 2026-02-09
Add SMU-Ossila_X200

franz-sweepMe:
The device has 2 channels for voltage sourcing and V/I measurements, which are implemented as an SMU driver. Auto-ranging is not supported by the hardware, but implemented in the driver.
The 2 dedicated V-measure channels VSense 1+2 are implemented as Logger drivers


FelixDollinger-SweepMe (dismissed):
Minor comment only. Looks good!

FelixDollinger-SweepMe (approved):
Minor changes.

PR #244, opened 2026-04-16
SMU -> add Undalogic miniSMU MS01 driver

joeltroughton:
Add driver for the miniSMU MS01 source measure unit.

### DIFF
--- a/src/SMU-Undalogic_miniSMU-MS01/main.py
+++ b/src/SMU-Undalogic_miniSMU-MS01/main.py
@@ -1,0 +1,676 @@
+# This Device Class is published under the terms of the MIT License.
+# Required Third Party Libraries, which are included in the Device Class
+# package for convenience purposes, may have a different license. You can
+# find those in the corresponding folders or contact the maintainer.
+#
+# MIT License
+#
+# Copyright (c) 2026 Undalogic Ltd
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+
+# SweepMe! driver
+# * Module: SMU
+# * Instrument: Undalogic miniSMU MS01
+
+from __future__ import annotations
+
+import json
+import re
+import time
+
+import numpy as np
+from pysweepme.EmptyDeviceClass import EmptyDevice
+
+# Current range indices mapped to their maximum current magnitude in amperes
+CURRENT_RANGE_LIMITS = {
+    0: 1e-6,    # Range 0: +/- 1 uA
+    1: 25e-6,   # Range 1: +/- 25 uA
+    2: 650e-6,  # Range 2: +/- 650 uA
+    3: 15e-3,   # Range 3: +/- 15 mA
+    4: 180e-3,  # Range 4: +/- 180 mA
+}
+
+
+class Device(EmptyDevice):
+    """Driver for the Undalogic miniSMU MS01 dual-channel source measure unit.
+
+    Supports voltage and current sourcing with measurement, hardware-accelerated
+    I-V sweeps, 4-wire Kelvin sensing, and both USB and WiFi connectivity.
+    """
+
+    description = """
+        <h3>Undalogic miniSMU MS01</h3>
+        <p>2-channel Source Measure Unit with FVMI/FIMV modes.</p>
+        <p><b>Connection:</b> USB (COM port) or WiFi (SOCKET port)</p>
+        <p><b>Features:</b></p>
+        <ul>
+            <li>Force Voltage / Measure Current (FVMI)</li>
+            <li>Force Current / Measure Voltage (FIMV)</li>
+            <li>Hardware-accelerated I-V sweep (up to 1000 points)</li>
+            <li>4-wire Kelvin sensing mode</li>
+            <li>5 current measurement ranges (1 uA to 180 mA)</li>
+            <li>Configurable oversampling (0-15)</li>
+        </ul>
+        <p>For WiFi connections, select the SOCKET port corresponding to the
+        device IP address and port (default TCP port: 3333).</p>
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        self.shortname = "miniSMU"
+
+        self.variables = ["Voltage", "Current"]
+        self.units = ["V", "A"]
+        self.plottype = [True, True]
+        self.savetype = [True, True]
+
+        self.port_manager = True
+        self.port_types = ["COM", "SOCKET"]
+        self.port_properties = {
+            "timeout": 1,
+            "baudrate": 115200,
+            "EOL": "\n",
+            # The miniSMU firmware TCP command parser uses strcmp() and does NOT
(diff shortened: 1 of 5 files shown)


src/SMU-Undalogic_miniSMU-MS01/main.py
```diff
@@ -0,0 +1,676 @@
...
+
+                if status == "COMPLETED":
+                    break
+                elif status == "ABORTED":
+                    msg = "Hardware I-V sweep was aborted."
+                    raise Exception(msg)
+                elif status == "RUNNING":
+                    time.sleep(0.1)
+                else:
+                    break
```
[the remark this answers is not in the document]
joeltroughton: Added a timeout based on the expected duration + 30s

src/SMU-Undalogic_miniSMU-MS01/main.py
```diff
@@ -0,0 +1,790 @@
...
+
+
+class Device(EmptyDevice):
+    """Driver for the Undalogic miniSMU MS01 dual-channel source measure unit.
+
+    Supports voltage and current sourcing with measurement, hardware-accelerated
+    I-V sweeps, 4-wire Kelvin sensing, and both USB and WiFi connectivity.
+    """
+
+    description = """
```
franz-sweepMe: The SweepMe! GUI displays the description saved in doc/description.md. The `description` parameter is only used as a fallback if no .md file is present and can be removed.

src/SMU-Undalogic_miniSMU-MS01/main.py
```diff
@@ -0,0 +1,790 @@
...
+                json.loads(current_response)
+                return current_response
+            except json.JSONDecodeError:
+                pass
+
+        # Read additional chunks until complete or timeout
+        max_timeout_iterations = 10
+        timeout_count = 0
+
+        while timeout_count < max_timeout_iterations:
```
franz-sweepMe, line 256: Whenever a driver runs a task that might take more than a few seconds or has the potential to cause the sequence to be stuck, it is advised to give the user the opportunity to stop the procedure. The `self.is_run_stopped()` function can be used to check if the user pressed the stop-button.

```suggestion
        while timeout_count < max_timeout_iterations and not self.is_run_stopped():
```

src/SMU-Undalogic_miniSMU-MS01/main.py
```diff
@@ -0,0 +1,790 @@
...
+
+    # --- Connection management -------------------------------------------------
+
+    def find_ports(self) -> list[str]:
+        """Provide a SOCKET port template with the default miniSMU TCP port.
+
+        The miniSMU firmware listens on TCP port 3333. This template lets the
+        user replace only the IP address without needing to remember the port.
+        """
+        return ["TCPIP0::<ip>::3333::SOCKET"]
```
franz-sweepMe: `self.port_types` was set to COM and SOCKET. The Port Manager will search for these two types of ports and a list of available ports will be displayed when pressing the "Find Ports" button. Additionally, this list will be extended by "TCPIP0::<ip>::3333::SOCKET" due to this function. 

If you want to display the TCPIP ports in the given format, you can set `self.port_types = ["COM", "TCPIP"]` directly. 
If you only want the user to see the template and not all other available TCPIP or Socket ports, you can remove "SOCKET" from the port types.
The port manager will receive the entered port string and automatically open the correct port type (COM or TCPIP), regardless of what is set in `self.port_types`

src/SMU-Undalogic_miniSMU-MS01/main.py
```diff
@@ -0,0 +1,790 @@
...
+            self._send_command(f"SOUR{ch}:SWEEP:AUTO:ENA")
+            # Use JSON format so the chunked JSON reader can accumulate the full
+            # multi-packet response.  CSV is line-delimited and the USB reader
+            # returns after the first line, which causes only 1 point to be stored.
+            self._send_command(f"SOUR{ch}:SWEEP:FORMAT JSON")
+
+    def poweron(self) -> None:
+        """Enable the output of the selected channel."""
+        self._send_command(f"OUTP{self.channel} ON")
+        self.output_on = True
```
franz-sweepMe: the `self.output_on` parameter is set but never used

src/SMU-Undalogic_miniSMU-MS01/main.py
```diff
@@ -0,0 +1,790 @@
...
+            # Execute hardware sweep
+            self._send_command(f"SOUR{self.channel}:SWEEP:EXECUTE")
+
+            # Poll for completion with timeout
+            # Timeout = generous estimate based on sweep parameters plus overhead
+            expected_duration_s = (self.listsweep_points * self.listsweep_dwell_ms) / 1000.0
+            timeout_s = expected_duration_s + 30.0
+            start_time = time.time()
+
+            while True:
```
franz-sweepMe, line 541:
```suggestion
            while not self.is_run_stopped():
```

src/SMU-Undalogic_miniSMU-MS01/main.py
```diff
@@ -0,0 +1,790 @@
...
+
+    def measure(self) -> None:
+        """Trigger a measurement or execute a hardware sweep.
+
+        In normal mode, sends an atomic voltage+current query.
+        In list sweep mode, executes the onboard sweep and polls until complete.
+        """
+        if self.use_list_sweep:
+            # Execute hardware sweep
+            self._send_command(f"SOUR{self.channel}:SWEEP:EXECUTE")
```
franz-sweepMe: The list sweep is started only for the selected channel. Does the device support parallel list sweeps for both channels?

src/SMU-Undalogic_miniSMU-MS01/main.py
```diff
@@ -0,0 +1,790 @@
...
+    def set_source_mode(self, channel: int, mode: str) -> None:
+        """Set the source mode on a channel.
+
+        Args:
+            channel: Channel number (1 or 2).
+            mode: Source mode ("FVMI" or "FIMV").
+        """
+        self._send_command(f"SOUR{channel}:{mode} ENA")
+
+    def set_sweep_parameters(
```
franz-sweepMe, line 629: Very nice that you included wrappers for all important functionalities. 
In the semantic functions (e.g. configure) you are sending the commands directly instead of using these wrappers. Consider using the wrapped functions so the commands can be maintained in one place

franz-sweepMe (changes requested):
Thank you for your contribution, this is a very well written driver!

You can set the comments to "resolved" when you implement the requested changes, or explain why the existing implementation is preferred. 
The comments by Gemini Code Assist can be seen as suggestions. If they are not applicable, you can set them to "resolved".
Once all comments are resolved, I am happy to merge and upload the driver to SweepMe!

src/SMU-Undalogic_miniSMU-MS01/main.py
```diff
@@ -0,0 +1,790 @@
...
+        # Read additional chunks until complete or timeout
+        max_timeout_iterations = 10
+        timeout_count = 0
+
+        while timeout_count < max_timeout_iterations:
+            try:
+                chunk = self.port.read()
+                if chunk:
+                    if chunk.strip() and self._is_valid_chunk(chunk):
+                        response_buffer.append(chunk)
```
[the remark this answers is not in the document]
joeltroughton: Removed the strip() guard. Every non-empty valid chunk is now appended so inter-chunk whitespace can't silently corrupt the reassembled JSON.

src/SMU-Undalogic_miniSMU-MS01/main.py
joeltroughton: Removed description

src/SMU-Undalogic_miniSMU-MS01/main.py
joeltroughton: Added `self.is_run_stopped()` checks in both long-running loops. The chunked-read timeout loop returns the buffered response early on stop. In the sweep polling loop, we additionally send `SOURx:SWEEP:ABORT` before returning so the device halts mid-sweep (otherwise a half-finished run can leave stale data in the onboard buffer and corrupt the next measurement run)

src/SMU-Undalogic_miniSMU-MS01/main.py
joeltroughton: Switched to `port_types = ["COM", "TCPIP"]`, dropped the find_ports template, and renamed the socket EOL keys to TCPIP_EOLwrite/TCPIP_EOLread/TCPIP_timeout.

src/SMU-Undalogic_miniSMU-MS01/main.py
joeltroughton: Removed

src/SMU-Undalogic_miniSMU-MS01/main.py
joeltroughton: Added `self.is_run_stopped()` check at the top of the polling loop in `measure()`.

src/SMU-Undalogic_miniSMU-MS01/main.py
joeltroughton: Yes, parallel sweeps should work. The firmware keeps per-channel sweep state and spawns independent tasks for each channel.

src/SMU-Undalogic_miniSMU-MS01/main.py
joeltroughton: Refactored functions to call the public helpers

franz-sweepMe (approved):

PR #247, opened 2026-04-23
B29xx -> read DC measurement in same phase as query

franz-sweepMe:
Fixes result buffer being overwritten when using multiple channel instances.


FelixDollinger-SweepMe (approved):
Small bugfix

PR #255, opened 2026-07-22
Undalogic miniSMU MS01 & uSMU: Config in configure(), sweep-start settling, robust list-sweep readout

joeltroughton:
Driver improvements for the Undalogic miniSMU MS01 and uSMU, addressing
SweepMe! semantic-function usage and measurement artifacts at sweep start.

## SMU-Undalogic_miniSMU-MS01

**Semantic function usage**
- Removed the unsupported `"TCPIP_timeout"` port property (the port manager
  warns about unknown keywords at run start).
- Moved voltage range, current range, oversampling, and compliance setup from
  `initialize()` into `configure()`, so values driven by the parameter syntax
  (e.g. a compliance limit from a GUI widget) are re-applied whenever they
  change during a running sequence. `initialize()` keeps the once-per-run
  work: validation, source mode (FVMI/FIMV), and 4-wire enable.

**List sweep: CSV instead of JSON**
- The onboard hardware sweep is now read back in CSV format (one
  `timestamp,voltage,current` line per point). Firmware up to 1.4.5 could
  drop a few bytes mid-stream when serializing large JSON responses, which
  made the whole document unparseable and the sweep unusable. CSV is
  line-framed, so a damaged row costs one point instead of the entire sweep.
  (The serializer itself is fixed in firmware 1.4.6; CSV remains the more
  robust choice either way.)

**Settling after large setpoint steps**
- The firmware autoranger steps down at most one current range per
  measurement, debounced over several readings. After a large setpoint jump —
  most visibly when a repeated sweep wraps from the last value back to the
  first, which SweepMe! performs with no semantic function call in between —
  the first points of the next pass reported the zero-current floor of a
  still-too-coarse range (a descending staircase from ~1e-5 A down to the
  true floor). The driver now detects setpoint jumps >= 0.5 V in `apply()`
  and discards readings until the autoranger has converged, and runs the same
  warm-up before every onboard list sweep. Costs roughly 0.2 s per sweep
  start on current firmware; normal sweep steps are unaffected.

## SMU-Undalogic_uSMU

- The output is driven to 0 V before enabling. The enable command only raises
  the amplifier-enable pin while the DAC keeps the last programmed voltage,
  so the previous run's final voltage appeared at the terminals from
  output-enable until the first sweep point was applied — visible as a brief
  flash when sweeping an LED.

## Testing

Bench-tested against a miniSMU MS01 over USB on firmware 1.4.6: SweepEditor
voltage sweeps (0 → 2.5 V, blue LED) with back-to-back repetitions now start
every pass at the true noise floor, and list-sweep mode returns complete,
parseable data across repeated runs. Works on firmware 1.4.5+; best results
with 1.4.6, which additionally fixes autorange-related jumps in the turn-on
region on the firmware side.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved measurement stability after large setpoint changes by allowing the instrument’s autoranging to settle before recording results.
  - Fixed list sweep data handling to reliably capture CSV measurements and avoid incomplete readings.
  - Ensured list sweeps start from the configured initial voltage.
  - The uSMU now sets output voltage to 0 V before enabling output, improving startup safety and consistency.

### DIFF
--- a/src/SMU-Undalogic_miniSMU-MS01/main.py
+++ b/src/SMU-Undalogic_miniSMU-MS01/main.py
@@ -132,6 +132,14 @@ class Device(EmptyDevice):
         self._sweep_currents: np.ndarray = np.array([])
         self._sweep_timestamps: np.ndarray = np.array([])
 
+        # After OUTP ON the source sits at 0; the first apply produces a large
+        # step that the firmware autoranger needs several MEAS calls to
+        # converge across (it advances at most one current range per call) and
+        # that the output stage needs time to slew through. Without an explicit
+        # settle pass on the first apply, the first 1-2 sweep points report
+        # clipped/transient values. Reset in poweron() so every run is fresh.
+        self._needs_first_apply_settle: bool = True
+
     # --- GUI interaction -------------------------------------------------------
 
     def set_GUIparameter(self) -> dict:
@@ -506,6 +514,7 @@ class Device(EmptyDevice):
     def poweron(self) -> None:
         """Enable the output of the selected channel."""
         self.enable_output(self.channel)
+        self._needs_first_apply_settle = True
 
     def poweroff(self) -> None:
         """Disable the output of the selected channel."""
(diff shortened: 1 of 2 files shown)


franz-sweepMe (changes requested):
Thank you for your contribution!
Please double-check Code Rabbits comment and ensure that the examples work with the updated driver.
Best,
Franz

